PR #7, opened 2016-09-27
Add JUCE's NetworkGraphicsDemo example

McMartin:
This PR allows passing any string to `jucer_project_begin()`, even if it contains characters that are not allowed in CMake target names. It also makes the string passed to `jucer_project_begin()` the name of the executable.


MartyLake:
Lgtm after reading! One fixup left before the pr is clean :-)

PR #26, opened 2017-01-20
Restructure CMake files

McMartin:
@MartyLake: in the end I decided to not keep the top-level `.cmake` file. We can still re-introduce it later, if we think it could make people's life easier. Please review, thanks!


McMartin:
@MartyLake: ping

PR #30, opened 2017-01-24
Pass project name like other settings

McMartin:

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -22,10 +22,12 @@
 set(Reprojucer.cmake_DIR "${CMAKE_CURRENT_LIST_DIR}")
 
 
-function(jucer_project_begin project_name)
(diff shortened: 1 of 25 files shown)


cmake/Reprojucer.cmake
```diff
@@ -51,7 +53,12 @@ function(jucer_project_begin project_name)
         )
       endif()
 
-      if(tag STREQUAL "PROJECT_VERSION")
+      if(POLICY CMP0054)
```
McMartin: No. `if(POLICY X)` is used to check that the current version of CMake knows about that policy. For instance, CMP0054 was introduced in CMake 3.1 and `cmake_policy(SET CMP0054 NEW)` would warn or fail on older versions.

MartyLake: OK I get it thanks ! LGTM then ^^

PR #54, opened 2017-04-27
Cpp jucer2 c make

MartyLake:


cmake/Jucer2CMake/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+  message(FATAL_ERROR "${Projucer_jucer_FILE} must exist")
+endif()
+
+# Explicitely makes Projucer file variable absolute, based the from build directory 
+# (which is the default base directory when calling the "NOT EXISTS" directive 
+# with a relative path).
+# This disambiguates further COPY operations (which are based on the source directory 
+# when called with a relative path).
+get_filename_component(Projucer_jucer_FILE "${Projucer_jucer_FILE}"
+                       REALPATH BASE_DIR "${CMAKE_BINARY_DIR}")
```
McMartin: @MartyLake: please try 
```
get_filename_component(Projucer_jucer_FILE "${Projucer_jucer_FILE}" ABSOLUTE)
```
and I guess it should go before checking for existence? Otherwise I don't understand what it's doing...

cmake/Jucer2CMake/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+
+get_filename_component(Projucer_jucer_DIR "${Projucer_jucer_FILE}" DIRECTORY)
+
+file(COPY "${Projucer_jucer_DIR}/Source/Utility/jucer_PresetIDs.h"
+  DESTINATION "${CMAKE_CURRENT_LIST_DIR}/Source/Utility"
+)
```
McMartin: @MartyLake: let's not copy `jucer_PresetIDs.h`, but instead add `${Projucer_jucer_DIR}/Source/Utility` as an include directory on `Jucer2CMake` (after the call to `jucer_project_end()`) and use `#include <jucer_PresetIDs.h>` in `main.cpp`.

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "JuceHeader.h"
+
+#include "Source/Utility/jucer_PresetIDs.h"
```
McMartin: @MartyLake: see https://github.com/McMartin/JUCE.cmake/pull/54/files#r113718620. Let's change this to `#include <jucer_PresetIDs.h>`

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+      juce::File::getCurrentWorkingDirectory().getFullPathName().toStdString();
+  const auto cmakeProjectJucerPrefix =
+      juce::File(jucerFilePath).getFileNameWithoutExtension().toStdString() +
+      "_jucer";
+  const auto cmakeProjectJucerFile = cmakeProjectJucerPrefix + "_FILE";
+  const auto cmakeProjectJucerDir = cmakeProjectJucerPrefix + "_DIR";
+  const auto cmakeJuceRootDir = "JUCE_ROOT";
+
+  DBG("Jucer file Path = " + jucerFilePath);
+  DBG("Reproducer Path = " + reproducerLibraryPath);
```
McMartin: @MartyLake: please remove debugging statements.

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+    return 1;
+  }
+
+  // Parse command line arguments
+  std::vector<std::string> args{argv, argv + argc};
+  const auto &jucerFilePath = args.at(1);
+  const auto jucerFolderPath = juce::File(jucerFilePath)
+                                   .getParentDirectory()
+                                   .getFullPathName()
+                                   .toStdString();
```
McMartin: @MartyLake: this variable is only used line 229, please move it there.

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+  const auto &reproducerLibraryPath = args.at(2);
+  const auto currentWorkingDirectory =
+      juce::File::getCurrentWorkingDirectory().getFullPathName().toStdString();
+  const auto cmakeProjectJucerPrefix =
+      juce::File(jucerFilePath).getFileNameWithoutExtension().toStdString() +
+      "_jucer";
```
McMartin: @MartyLake: this variable should contain a string that can be used as a CMake variable, i.e. it's not enough to replace `.jucer` by `_jucer`, there might be other characters that are invalid. I think JUCE has a function to convert a string into a valid C++ identifier, we should use that instead.

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+  const auto cmakeProjectJucerFile = cmakeProjectJucerPrefix + "_FILE";
```
McMartin: @MartyLake: this variable is not necessary, `_FILE` should be printed directly.

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+  const auto cmakeProjectJucerDir = cmakeProjectJucerPrefix + "_DIR";
```
McMartin: @MartyLake: same thing for this variable, `_DIR` should be printed directly.

cmake/Jucer2CMake/CMakeLists.txt
MartyLake: It's not, for the reasons written in the comment. I'll move it before existence checking.

cmake/Jucer2CMake/CMakeLists.txt
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: @McMartin The closest I found is Identifier::isValidIdentifier :

Since Identifiers are used as a script variables and XML attributes, they should only contain alphanumeric characters, underscores, or the '-' and ':' characters.

Is it enough for us ?

cmake/Jucer2CMake/main.cpp
MartyLake: 👍

cmake/Jucer2CMake/CMakeLists.txt
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: Ok, I imported the method from projucer jucer_CodeHelpers.cpp::l54::makeValidIdentifier

MartyLake: :+1:

cmake/Jucer2CMake/CMakeLists.txt
McMartin: Ok, I understand why we need to make it absolute from `${CMAKE_BINARY_DIR}`, but please reword the comment, I find it harder to understand with the comment than without.

McMartin: Also the comment still refers to COPY operations, though we don't have any anymore.

cmake/Jucer2CMake/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+
+jucer_project_module(
+  juce_gui_extra
+  PATH "${Projucer_jucer_DIR}/../../modules"
+)
+
+jucer_project_end()
+
+
+include_directories("${Projucer_jucer_DIR}/Source/Utility/")
```
McMartin: @MartyLake: please use `target_include_directories`

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if (CharacterFunctions::isLetter(s[i]) &&
+        CharacterFunctions::isLetter(s[i - 1]) &&
+        CharacterFunctions::isUpperCase(s[i]) &&
+        !CharacterFunctions::isUpperCase(s[i - 1]))
+      s = s.substring(0, i) + " " + s.substring(i);
+
+  String allowedChars(
+      "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_ 0123456789");
+  if (allowTemplates)
+    allowedChars += "<>";
```
McMartin: @MartyLake: there are no templates in CMake

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  if (!removeColons)
+    allowedChars += ":";
```
McMartin: @MartyLake: I don't think colon is a valid character in a CMake identifier

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,302 @@
...
+      n << words[i].substring(0, 1).toUpperCase()
+        << words[i].substring(1).toLowerCase();
+    else
+      n << words[i];
+  }
+
+  if (CharacterFunctions::isDigit(n[0]))
+    n = "_" + n;
+
+  if (CPlusPlusCodeTokeniser::isReservedKeyword(n))
```
McMartin: @MartyLake: we don't really care about reserved keyword in C++.

cmake/Jucer2CMake/main.cpp
```diff
@@ -0,0 +1,302 @@
...
+  // Parse command line arguments
+  std::vector<std::string> args{argv, argv + argc};
+  const auto &jucerFilePath = args.at(1);
+  const auto &reproducerLibraryPath = args.at(2);
+  const auto currentWorkingDirectory =
+      juce::File::getCurrentWorkingDirectory().getFullPathName().toStdString();
+  const auto cmakeProjectJucerPrefix =
+      makeValidIdentifier(
+          juce::File(jucerFilePath).getFileNameWithoutExtension() + "_jucer",
+          false, true, false)
```
McMartin: @MartyLake: instead of passing `false, true, false`, let's change the function to not take these parameters.

McMartin:
@MartyLake: f397576d77950a723a1d3ff357136bd7abf06b9a should be removed. We need to check whether `Projucer_jucer_FILE` is defined before doing anything with it.

cmake/Jucer2CMake/CMakeLists.txt
MartyLake: :+1:

cmake/Jucer2CMake/CMakeLists.txt
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
MartyLake: :+1:

cmake/Jucer2CMake/main.cpp
```diff
@@ -36,36 +36,16 @@ void printError(const std::string &error) {
...
-
-  s = s.replaceCharacters(".,;:/@", "______");
-
-  for (int i = s.length(); i > 0; --i) {
-    if (CharacterFunctions::isLetter(s[i]) &&
-        CharacterFunctions::isLetter(s[i - 1]) &&
-        CharacterFunctions::isUpperCase(s[i]) &&
-        !CharacterFunctions::isUpperCase(s[i - 1])) {
-      s = s.substring(0, i) + " " + s.substring(i);
+  auto stdString = s.toStdString();
```
McMartin: @MartyLake: there are too many back-and-forth between `juce::String` and `std::string` for my taste. Let's try to use `juce::String` as less as possible.

cmake/Jucer2CMake/main.cpp
```diff
@@ -20,7 +20,7 @@
 // SOFTWARE.
 
 #include "JuceHeader.h"
-
+typedef juce::Identifier Identifier;
```
McMartin: @MartyLake: `using juce::Identifier;` would be more correct

cmake/Jucer2CMake/main.cpp
```diff
@@ -64,20 +83,20 @@ int main(int argc, char *argv[]) {
...
+      juce::File(jucerFilePath).getFileNameWithoutExtension().toStdString() +
+      "_jucer");
   const auto cmakeJuceRootDir = "JUCE_ROOT";
 
   // Load jucer project
-  auto file = File{jucerFilePath};
-  ScopedPointer<XmlElement> xml = XmlDocument::parse(file);
+  auto file = juce::File{jucerFilePath};
+  std::unique_ptr<juce::XmlElement> xml;
+  xml.reset(juce::XmlDocument::parse(file));
```
McMartin: @MartyLake: `auto xml = std::unique_ptr<juce::XmlElement>{juce::XmlDocument::parse(file)};`

PR #55, opened 2017-04-28
Reset JUCE to the last commit of version 4.3.1

McMartin:
JUCE.cmake doesn't support JUCE 5 yet.


MartyLake:
LGTM

PR #57, opened 2017-04-28
Respect the terms of the GNU General Public License

McMartin:
@MartyLake: please review and confirm that you agree with the change of license for the files that you authored, thanks!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,674 @@
-The MIT License (MIT)
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-Copyright (c) 2016 Alain Martin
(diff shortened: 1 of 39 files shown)


MartyLake:
LGTM. I agree with the change of license for the files I authored !

MartyLake:
Do we have to copyright every source file ? Or just the repository ?

MartyLake:
I think we have to keep it this way: https://www.gnu.org/licenses/gpl-faq.html#GPLOmitPreamble

McMartin:
According to https://github.com/McMartin/JUCE.cmake/blob/97ce486c4159c283c223b13f15c854d374b5797f/LICENSE#L629, it is safer to do so.

PR #97, opened 2017-05-08
Fix parsing .projucer files with only one exporter

gonzaloflirt:
Use the contents of the first exporter when parsing the modulepath.


McMartin:
@gonzaloflirt: Thanks!

PR #99, opened 2017-05-11
Handle preprocessor definitions written one per line

McMartin:

### DIFF
--- a/Jucer2CMake/CMakeLists.txt
+++ b/Jucer2CMake/CMakeLists.txt
@@ -40,6 +40,7 @@ jucer_project_begin(
   PROJECT_NAME "Jucer2CMake"
   PROJECT_VERSION "1.0.0"
   PROJECT_TYPE "Console Application"
+  PREPROCESSOR_DEFINITIONS "DONT_SET_USING_JUCE_NAMESPACE=1"
 )
 
 jucer_project_files("Jucer2CMake"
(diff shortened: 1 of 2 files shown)


MartyLake:
LGTM. 

however, wouldn't it be easier to std:replace instead of join + split?

McMartin:
@MartyLake: Good point. Also we are not escaping existing `;`. I guess I should write something less generic, and more geared towards writing CMake code.

MartyLake:
@McMartin

Maybe the join  function can take a list of characters to escape as another
 parameter, and Espace any character that exists in any of the strings
passed in?

McMartin:
@MartyLake: I added two `escape` functions (one for `std::string` and one for `std::vector<std::string>`) so we can combine calls to `split`, `escape` and `join`. Please have another look!

MartyLake:
Besides the minor correction, looks good to me!

MartyLake:
Good job !!! :)

McMartin:
@MartyLake: this PR doesn't work... 😞 
When calling
```
jucer_project_begin(
  PROJECT_NAME "Foo"
  PROJECT_TYPE "GUI Application"
  PREPROCESSOR_DEFINITIONS "A=1;B=2"
)
```
we get the error message: `Unsupported project setting: B=2`.
I need to figure out another solution...

McMartin:
Fixed by https://github.com/McMartin/JUCE.cmake/pull/103

PR #117, opened 2017-05-18
Warn when jucer_export_target() was called with non-relevant exporters

McMartin:


MartyLake:
LGTM!

McMartin:
@MartyLake: the first four commits of this PR are now identical to the first four commits of #118. Please review both PRs and tell me which one you prefer, thanks!

PR #155, opened 2017-05-29
Fix setting CMAKE_OSX_SYSROOT when using the Make generator

McMartin:
@MartyLake: please test with the Make generator and review, thanks!


MartyLake:
LGTM

PR #181, opened 2017-06-06
Replace the list of supported commits by a function

McMartin:


McMartin:
On second thoughts, I am not really happy about this PR. I'll reopen something similar later.

McMartin:
In the end we decided to use a simpler approach based on the `jucerVersion` property of the `.jucer` project (see https://github.com/McMartin/JUCE.cmake/pull/232).

PR #223, opened 2017-07-22
Fix setting target properties that depend on $<CONFIG>

McMartin:

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -1954,14 +1954,17 @@ function(__set_plugin_output_directory_property
   target_name plugin_type plugins_dir plugin_extension
 )
 
+  set(default_output_dir "$ENV{HOME}/Library/Audio/Plug-Ins/${plugins_dir}")
+
   foreach(configuration_name ${JUCER_PROJECT_CONFIGURATIONS})
     if(DEFINED JUCER_${plugin_type}_BINARY_LOCATION_${configuration_name})
       set(output_dir ${JUCER_${plugin_type}_BINARY_LOCATION_${configuration_name}})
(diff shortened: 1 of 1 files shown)


cmake/Reprojucer.cmake
```diff
@@ -1554,6 +1554,7 @@ function(__set_common_target_properties target_name)
     endif()
     string(APPEND all_confs_output_name $<$<CONFIG:${configuration_name}>:${output_name}>)
   endforeach()
+  string(APPEND all_confs_output_name $<$<CONFIG:>:${JUCER_PROJECT_NAME}>)
```
MartyLake: Should this "advanced" trick inlined in code, instead of the commit's message ?

McMartin: You want me to add a comment?

MartyLake:
Maybe a comment would be better suited than a long commit message ? IDK


---
Matthieu TALBOT
neogaldr@gmail.com

+33 629 22 15 70
fr.linkedin.com/in/matthieutalbot
http://www.soundcloud.com/Marty-Lake
http://lifehacker.com/5028808/how-eom-makes-your-email-more-efficient

McMartin:
I took quite some time to write the commit message. Don't you like it? :crying_cat_face: 
More seriously, I'll add a comment before merging. Thanks for the review!

MartyLake:
I love it! I'm more afraid that no one will ever stumble upon it, which
would be a shame! ;-)

McMartin:
@MartyLake: I added a simple comment. If people want to know more, they can look at the git history.

MartyLake:
that's great ! :)

PR #225, opened 2017-07-25
Do not use Reprojucer.cmake to configure its helper executables

McMartin:

### DIFF
--- a/cmake/BinaryDataBuilder/CMakeLists.txt
+++ b/cmake/BinaryDataBuilder/CMakeLists.txt
@@ -18,10 +18,6 @@
 cmake_minimum_required(VERSION 3.4)
 
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../../cmake")
-include(Reprojucer)
-
-
 if(NOT DEFINED JUCE_modules_DIRS)
   message(FATAL_ERROR "JUCE_modules_DIRS must be defined")
 endif()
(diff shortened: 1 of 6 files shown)


McMartin:
@MartyLake: The diff is not really readable, so it might be easier to have a look at the final state directly:
- https://github.com/McMartin/JUCE.cmake/blob/a54ebbeb91bd1dca279538f8886a2b0ec5ea77d1/cmake/BinaryDataBuilder/CMakeLists.txt
- https://github.com/McMartin/JUCE.cmake/blob/a54ebbeb91bd1dca279538f8886a2b0ec5ea77d1/cmake/IconBuilder/CMakeLists.txt

MartyLake:
Since one JUCE_modules_DIR must contain **all** juce folders ( https://github.com/McMartin/JUCE.cmake/blob/a54ebbeb91bd1dca279538f8886a2b0ec5ea77d1/cmake/BinaryDataBuilder/CMakeLists.txt#L26 ), it doesn't make sense to allow the user to provide a list, does it ?

McMartin:
The main (and maybe only) user is `Reprojucer.cmake` and being able to provide a list makes its code simpler (see https://github.com/McMartin/JUCE.cmake/blob/master/cmake/Reprojucer.cmake#L1434). Anyway, this is the only thing that didn't change in that PR (and I don't intend to change it now). :smile:

MartyLake:
LGTM then ;)

PR #247, opened 2017-10-10
Fix bug reported in issue #246

### LINKED ISSUE #246: Problem with prebuild script and post build script on Windows
Hi @McMartin,
It seems that there is a problem with the prebuild and postbuild scripts on the Windows Toolchain, in Release Mode. The symptom is an invalid command line in the vcxproj file in the pre/post build script field:
`"" <path>\build\Release\postbuild_Release.cmd`
instead of
`call <path>\build\Release\postbuild_Release.cmd`.
We looked for the commit which trigger this bug, and found that [this commit]( https://github.com/McMartin/JUCE.cmake/commit/a443911e2f98f12860538c2a3d1318c7a7783f93) is responsible for the behavior. Especially the changes on the lines 1755 and 1809. Reversing the changes on those lines seems to fix the bug.

McMartin:
Resolves #246 

@MartyLake, @Xav83: please review, thanks!

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -2224,14 +2224,14 @@ function(__set_common_target_properties target)
 
       if(DEFINED JUCER_PREBUILD_COMMAND_${config})
         set(prebuild_command ${JUCER_PREBUILD_COMMAND_${config}})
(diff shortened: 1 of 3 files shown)


MartyLake:
Lgtm

PR #254, opened 2017-10-14
Catch target name mistakes

McMartin:


MartyLake:
I think this one need some explanations to me

McMartin:
@MartyLake: I added a commit at the beginning of the branch, and extended the last commit message. Please have another look, thanks!

PR #257, opened 2017-10-14
 Use more idiomatic case for module variables

McMartin:


McMartin:
On second thought, it doesn't make sense to treat these variables as module variables and expose them like this. We should rather "hide" them as much as possible (by wrapping them into private functions for instance).

Issue #292, opened 2017-12-18
Link issue with JUCE 5.2.0


Xav83:
Hi @McMartin !

We've recently trying to upgrade the JUCE version on our project from the version 4.3 to the version 5.2.0, and we have encountered a problem on Windows, with Visual Studio 2015.
In our project, we are building the vst and the vst3 version of our plugin, everything is compiling fine, but during the linking, an error occurs :
`include_juce_audio_plugin_client_VST3.obj : error LNK2005: _DllMain@12 already defined in include_juce_audio_plugin_client_VST2.obj`

We have tried to use the last version of the VST sdk, and then, after seing this [thread on the juce forum](https://forum.juce.com/t/no-flock-cpp-in-vst-sdk/25069/10), we have tried the sdk you've linked, without success. We had either the error described above, or some errors because of missing files that Juce is looking for while VST sdk version has moved or removed those files.

When building the project without FRUT, with Projucer, this error does not occur.

Thanks in advance for your attention, and if you need any other information about this problem, just ask 😉

Xav83:
I forgot to mention, we've also updated the FRUT version to the current master ! 😃

McMartin:
@Xav83 as usual, thanks a lot for the detailed bug report.

You are hitting the limitations of FRUT wrt. using JUCE 5... As mentioned on the README (https://github.com/McMartin/FRUT#requirements), JUCE 5 support is still experimental.

I am working on a branch that improves the JUCE 5 support and should fix you issue. What is your ideal time frame to switch to JUCE 5 (if you reply "yesterday", I'll try to hack down something :smile:)

Xav83:
@McMartin You're welcome.

I knew that JUCE 5 support is still experimental, but I wanted to see if our project will work with it. 
FYI, we haven't faced any problem on OSX, well done ! 👍 
I've created this issue to be able to give you some feedback about this update attempt 😉 

Haha If you can make it happen yesterday it will be perfect 😆 
Jokes aside, don't worry, if we'd really need to go to Juce 5, we could still generate our project with the Projucer and `xcodebuild` to automate it, and switch back to FRUT, when the JUCE 5 support will be more advanced, so there is no need to rush.

If you ever need some feedback on JUCE 5 support, you can ask us (@WGuLL or myself), we will update FRUT on our repository and tell you what's working and what's broken 😉

PR #293, opened 2017-12-18
Support the "Build Standalone Plug-In" project setting

McMartin:

@Xav83 this should fix https://github.com/McMartin/FRUT/issues/292, please give it a try, thanks!


Xav83:
@McMartin This PR fix indeed the bug described in #292 ! GGWP ! 
And thanks for the quick fix ! 😄

McMartin:
@Xav83 thanks for testing the PR!

@lethal-guitar feel free to do a quick review, since @MartyLake is on holiday, thanks!

cmake/Reprojucer.cmake
```diff
@@ -1781,7 +1812,11 @@ function(__generate_AppConfig_header)
     __bool_to_int("${JUCER_BUILD_AUDIOUNIT_V3}" Build_AUv3_value)
     __bool_to_int("OFF" Build_RTAS_value) # Not yet supported
     __bool_to_int("OFF" Build_AAX_value) # Not yet supported
-    __bool_to_int("${JUCER_BUILD_AUDIOUNIT_V3}" Build_STANDALONE_value)
+    if(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)
```
lethal-guitar: Would it make sense to define a global variable somewhere with this expression? You're checking it in quite a few places now.

cmake/Reprojucer.cmake
```diff
@@ -345,19 +351,26 @@ function(jucer_project_module module_name PATH_TAG modules_folder)
...
     )
       set(to_compile FALSE)
     endif()
 
+    if(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)
+      if(src_file MATCHES "_Standalone[._]" AND NOT (JUCER_BUILD_AUDIOUNIT_V3 AND APPLE))
+        set(to_compile FALSE)
+      endif()
+    else()
+      if(src_file MATCHES "_Standalone[._]" AND NOT JUCER_BUILD_STANDALONE_PLUGIN)
```
lethal-guitar: You could combine the two `if` statements into `(!(audio_unit_v3 && APPLE) || !(jucer_build_standalone_plugin)`

cmake/Reprojucer.cmake
```diff
@@ -1622,33 +1635,51 @@ function(jucer_project_end)
...
+      endif()
+
+      if(juce4_standalone OR (JUCER_BUILD_STANDALONE_PLUGIN
+          AND NOT (DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)))
+        if(juce4_standalone)
+          set(standalone_target ${target}_AUv3_Standalone)
+        else()
+          set(standalone_target ${target}_StandalonePlugin)
+        endif()
+        add_executable(${standalone_target} WIN32 MACOSX_BUNDLE
```
lethal-guitar: `WIN32 MACOSX_BUNDLE`? Does that make sense?

cmake/Reprojucer.cmake
```diff
@@ -2694,32 +2694,30 @@ endfunction()
...
-          if(NOT CMAKE_MATCH_0)
-            message(FATAL_ERROR "Invalid Xcode flag: \"${xcode_flag}\"")
-          endif()
-          list(APPEND all_flags "${CMAKE_MATCH_1}")
-          set(value "${CMAKE_MATCH_2}")
-          string(APPEND all_confs_${CMAKE_MATCH_1} "$<$<CONFIG:${config}>:${value}>")
-        endforeach()
-      endif()
-    endforeach()
+  set(all_flags)
```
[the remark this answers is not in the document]
McMartin: Good idea! I'll reword 03dcf97 to `Don't restrict __set_custom_xcode_flags() to macOS`

cmake/Reprojucer.cmake
McMartin: I cannot do that, since the condition on `JUCER_VERSION` matters. In JUCE 4, files that match `_Standalone[._]` should be compiled only when the user has enabled `Build AudioUnit v3`, but in JUCE 5, these files should be compiled only when the user has enabled `Build Standalone Plug-In`.

cmake/Reprojucer.cmake
McMartin: Yes. It means that this executable is a GUI application. `WIN32` will be translated to the linker flag `/SUBSYSTEM:WINDOWS` on Windows, and `MACOSX_BUNDLE` will produce an `.app` bundle on macOS. These settings don't have any impact on other platforms.

cmake/Reprojucer.cmake
```diff
@@ -2082,13 +2082,21 @@ function(__generate_icon_file icon_format out_icon_filename)
...
 
 
-function(__set_output_directory_properties target)
+function(__set_output_directory_properties target subfolder)
 
   foreach(config ${JUCER_PROJECT_CONFIGURATIONS})
     string(TOUPPER "${config}" upper_config)
 
-    if(DEFINED JUCER_BINARY_LOCATION_${config})
+    if(MSVC AND NOT (DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5))
```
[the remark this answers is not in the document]
McMartin: Indeed. When `JUCER_VERSION` is undefined, it means "latest", and the behavior of the latest Projucer (5.2.0) is to add sub-directories.

cmake/Reprojucer.cmake
```diff
@@ -2082,13 +2082,21 @@ function(__generate_icon_file icon_format out_icon_filename)
...
+      if(DEFINED JUCER_BINARY_LOCATION_${config})
+        set(output_directory "${JUCER_BINARY_LOCATION_${config}}/${subfolder}")
+      else()
+        set(output_directory "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/${subfolder}")
```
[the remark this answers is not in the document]
McMartin: `$<CONFIG>` is a [generator expression](https://cmake.org/cmake/help/v3.4/manual/cmake-generator-expressions.7.html#cmake-generator-expressions-7), which will be evaluated during the generation of the build system. I could use `${config}` here as well, but `$<CONFIG>` guarantees that the folder name is exactly the configuration name as shown in the IDE.

cmake/Reprojucer.cmake
```diff
@@ -2082,13 +2082,21 @@ function(__generate_icon_file icon_format out_icon_filename)
...
+      endif()
+    elseif(DEFINED JUCER_BINARY_LOCATION_${config})
       set(output_directory "${JUCER_BINARY_LOCATION_${config}}")
+    endif()
+    if(output_directory)
       set_target_properties(${target} PROPERTIES
```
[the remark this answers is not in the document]
McMartin: Good catch! There is definitively a risk that `output_directory` is still defined, since this code is in a `foreach` loop. A previous iteration of the loop could have defined `output_directory` and then we would be using it wrongly. Fixed in ee2d689.

cmake/Reprojucer.cmake
McMartin: I think there are already too many global variables, so I want to avoid adding more if possible. I could write
```
if(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)
  set(juce4 TRUE)
else()
  set(juce4 FALSE)
endif()
```
in functions that use that condition many times, but I won't do it in this PR.

cmake/Reprojucer.cmake
```diff
@@ -2082,13 +2082,21 @@ function(__generate_icon_file icon_format out_icon_filename)
...
-function(__set_output_directory_properties target)
+function(__set_output_directory_properties target subfolder)
 
   foreach(config ${JUCER_PROJECT_CONFIGURATIONS})
     string(TOUPPER "${config}" upper_config)
 
-    if(DEFINED JUCER_BINARY_LOCATION_${config})
+    if(MSVC AND NOT (DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5))
+      if(DEFINED JUCER_BINARY_LOCATION_${config})
+        set(output_directory "${JUCER_BINARY_LOCATION_${config}}/${subfolder}")
```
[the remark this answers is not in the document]
McMartin: Not sure... In doubt I added an extra condition in 6fbd017.

cmake/Reprojucer.cmake
```diff
@@ -1605,15 +1605,17 @@ function(jucer_project_end)
         __link_osx_frameworks(${auv3_target} "AudioUnit" "CoreAudioKit" "AVFoundation")
         __add_xcode_resources(${auv3_target})
         __set_custom_xcode_flags(${auv3_target})
+        unset(auv3_target)
+      endif()
 
-        # AUv3 Standalone
+      if(TARGET ${target}_AUv3_AppExtension)
```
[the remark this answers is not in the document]
McMartin: In JUCE/Projucer 4, when users enable `Build AudioUnit v3` they get 2 targets: the AUv3 AppExtension and the AUv3 Standalone (an app that loads the extension). In JUCE/Projucer 5, the Standalone target is available on all platforms, and is not dependent on `Build AudioUnit v3` anymore on macOS.

The point of d8f4ea6 (`Make building the Standalone target depend on building the AUv3 one`) is to prepare to 6b361d6 (`Support the "Build Standalone Plug-In" project setting`) by changing the condition under which we add the `*_AUv3_Standalone` target. The behavior doesn't change, but the code is written in a way that makes the diff of 6b361d6 smaller.

I guess I could reword the commit message to 
```
Unset auv3_target in preparation for the next-but-one commit

`${target}_AUv3_AppExtension` is added when `JUCER_BUILD_AUDIOUNIT_V3 AND APPLE`
is true, but in JUCE 5 the Standalone Plugin target can be added when
`Build AudioUnit v3` is not enabled and on other platforms than macOS.
```

McMartin:
@lethal-guitar thanks a lot for the partial review and the comments! I pushed 2 fixups and I will reword 2 commits. I am looking forward to reading the end of your review.

cmake/Reprojucer.cmake
lethal-guitar: Ah yes, I see it now..

cmake/Reprojucer.cmake
lethal-guitar: Sounds reasonable!

cmake/Reprojucer.cmake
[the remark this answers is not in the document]
lethal-guitar: Ah, interesting! :+1:

cmake/Reprojucer.cmake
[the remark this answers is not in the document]
lethal-guitar: Ok, understood now! The proposed additional explanation in the commit message sounds good, although I would suggest:

> Separate code paths for AUv3 targets

For the commit message itself. What do you say?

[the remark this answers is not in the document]
McMartin: That sounds really good. Thanks!

cmake/Reprojucer.cmake
```diff
@@ -345,19 +351,26 @@ function(jucer_project_module module_name PATH_TAG modules_folder)
   foreach(src_file ${module_src_files})
     unset(to_compile)
 
-    # See LibraryModule::CompileUnit::isNeededForExporter()
```
lethal-guitar: Was the comment intentionally removed?

cmake/Reprojucer.cmake
```diff
@@ -1622,33 +1635,51 @@ function(jucer_project_end)
...
-      if(TARGET ${target}_AUv3_AppExtension)
-        set(standalone_target ${target}_AUv3_Standalone)
-        add_executable(${standalone_target} MACOSX_BUNDLE
+      if(JUCER_BUILD_AUDIOUNIT_V3
+          AND DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)
+        set(juce4_standalone ON)
+      endif()
+
+      if(juce4_standalone OR (JUCER_BUILD_STANDALONE_PLUGIN
+          AND NOT (DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)))
```
lethal-guitar: Could you make using `JUCER_BUILD_STANDALONE_PLUGIN` an error when using Juce 4 (it's already a warning now)? You wouldn't need this `AND NOT (DEFINED ...)` part of the expression here.

cmake/Reprojucer.cmake
```diff
@@ -2640,11 +2675,13 @@ endfunction()
 
 function(__set_JucePlugin_Build_defines target target_type)
 
-  # See XCodeProjectExporter::Target::getTargetSettings()
```
lethal-guitar: Same question here about the comments

cmake/Reprojucer.cmake
```diff
@@ -2640,11 +2675,13 @@ endfunction()
...
-  # in JUCE/extras/Projucer/Source/Project Saving/jucer_ProjectExport_XCode.h
-  set(plugin_types     VST VST3 AudioUnit AudioUnitv3  RTAS AAX Standalone  )
-  set(setting_suffixes VST VST3 AUDIOUNIT AUDIOUNIT_V3 RTAS AAX AUDIOUNIT_V3)
-  set(define_suffixes  VST VST3 AU        AUv3         RTAS AAX Standalone  )
+  if(JUCER_BUILD_AUDIOUNIT_V3 AND DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5)
+    set(JUCER_BUILD_STANDALONE_PLUGIN ON)
+  endif()
+
+  set(plugin_types     VST VST3 AudioUnit AudioUnitv3  RTAS AAX Standalone       )
+  set(setting_suffixes VST VST3 AUDIOUNIT AUDIOUNIT_V3 RTAS AAX STANDALONE_PLUGIN)
```
lethal-guitar: What is this doing?

cmake/Reprojucer.cmake
McMartin: Yes. The code moved a bit between JUCE 4 and JUCE 5, making the comment not completely accurate.

cmake/Reprojucer.cmake
McMartin: Same answer. The code moved and the comment is not accurate anymore.

cmake/Reprojucer.cmake
McMartin: I could, but I don't really want to. I want to allow users to go back and forth between JUCE 4 and JUCE 5 if they want, without FRUT getting in their way.

cmake/Reprojucer.cmake
McMartin: These 3 lists (`plugin_types`, `setting_suffixes`, and `define_suffixes`) are used in the following `foreach` loop, where we iterate over them at the same time (with an index going from 0 to 6 included). I could have unrolled that `foreach` loop and written the "same" code 7 times, but using these lists is more compact. The function is used to set the 7 `JucePlugin_Build_*` defines to either 1 or 0. I could also re-write the code to look more like what Projucer does: https://github.com/WeAreROLI/JUCE/blob/master/extras/Projucer/Source/ProjectSaving/jucer_ProjectExporter.cpp#L411-L435 (but not in this PR).

McMartin: And the diff is dealing with the fact that in JUCE 4, the Standalone plugin is depending on the AUv3 one, while it's not the case anymore in JUCE 5.

McMartin:
@lethal-guitar Thank you so much for the review! It's time to rebase, reword and apply the fixups.

PR #294, opened 2017-12-22
Support the "Header search paths" project setting

McMartin:
@lethal-guitar. @MartyLake: please review, thanks!


cmake/Reprojucer.cmake
```diff
@@ -183,6 +184,16 @@ function(jucer_project_settings)
       elseif(tag STREQUAL "PREPROCESSOR_DEFINITIONS")
         string(REPLACE "\n" ";" value "${value}")
 
+      elseif(tag STREQUAL "HEADER_SEARCH_PATHS")
+        string(REPLACE "\\" "/" value "${value}")
+        string(REPLACE "\n" ";" value "${value}")
```
lethal-guitar: Does this work correctly with all types of line endings, if someone writes a `CMakeLists.txt` directly without using the converter?

cmake/Reprojucer.cmake
```diff
@@ -183,6 +184,16 @@ function(jucer_project_settings)
...
+        unset(header_search_paths)
```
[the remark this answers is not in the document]
McMartin: I considered it but decided against, since it matches the same code in `jucer_export_target_configuration`: https://github.com/McMartin/FRUT/blob/3db7295ad0b78dcd1c690eaa65c085fd81d4bc97/cmake/Reprojucer.cmake#L830-L833

cmake/Reprojucer.cmake
McMartin: CMake takes care of treating `\r\n` like `\n`, so we don't have to think about it.

PR #295, opened 2017-12-22
Don't restrict the "Company Copyright" project setting to JUCE 5.2.0

McMartin:

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -1220,7 +1220,8 @@ function(jucer_project_end)
   endif()
 
   if(WIN32 AND NOT JUCER_PROJECT_TYPE STREQUAL "Static Library")
-    if(NOT(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5.2))
+    if(DEFINED JUCER_COMPANY_COPYRIGHT
+        OR NOT(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5.2))
       set(resources_rc_legal_copyright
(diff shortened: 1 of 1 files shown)


cmake/Reprojucer.cmake
```diff
@@ -1220,7 +1220,7 @@ function(jucer_project_end)
   endif()
 
   if(WIN32 AND NOT JUCER_PROJECT_TYPE STREQUAL "Static Library")
-    if(NOT(DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5.2))
+    if(DEFINED JUCER_COMPANY_COPYRIGHT)
```
lethal-guitar: Who is defining `JUCER_COMPANY_COPYRIGHT`?

lethal-guitar:
The new version of the code looks good, I don't know how it worked previously. Would jucer_copyright always be defined when using JUCE 5.2 or newer?

cmake/Reprojucer.cmake
McMartin: It's done by `jucer_project_settings()`  https://github.com/McMartin/FRUT/blob/1487c70dd613a05d1aa0cb3cbca130c5b08d8c42/cmake/Reprojucer.cmake#L199

when `tag` equals `COMPANY_COPYRIGHT`.

McMartin:
@lethal-guitar I rebased the branch on the latest `master` (after merging #296) and I changed the code to be more accurate. Now `JUCER_COMPANY_COPYRIGHT` is used either when it's defined, or when using JUCE 5.2.0 and later.

PR #296, opened 2017-12-22
Simplify the condition to set JucePlugin_Build_* defines

McMartin:

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -2716,8 +2716,8 @@ function(__set_JucePlugin_Build_defines target target_type)
     list(GET plugin_types ${index} plugin_type)
     list(GET define_suffixes ${index} define_suffix)
 
-    if(JUCER_BUILD_${setting_suffix} AND (target_type STREQUAL "SharedCodeTarget"
-        OR target_type STREQUAL "${plugin_type}PlugIn"))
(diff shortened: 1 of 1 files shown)


cmake/Reprojucer.cmake
```diff
@@ -2716,8 +2716,8 @@ function(__set_JucePlugin_Build_defines target target_type)
     list(GET plugin_types ${index} plugin_type)
     list(GET define_suffixes ${index} define_suffix)
 
-    if((JUCER_BUILD_${setting_suffix} AND target_type STREQUAL "SharedCodeTarget")
-        OR target_type STREQUAL "${plugin_type}PlugIn")
+    if(target_type STREQUAL "${plugin_type}PlugIn"
+        OR (JUCER_BUILD_${setting_suffix} AND target_type STREQUAL "SharedCodeTarget"))
```
lethal-guitar: I would suggest also changing the order in the 2nd expression, so that it reads `target_type STREQUAL "SharedCodeTarget" AND ...`. This makes both operands to the `OR` start with a comparison on  `target_type`, making it nicely symmetric.

McMartin: Really good idea! Done in 1da7b06f93a06ba4070a6c7c9777597ee5f48e77.

PR #302, opened 2018-01-10
Fix two bugs in Jucer2Reprojucer

McMartin:

### DIFF
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -370,6 +370,9 @@ int main(int argc, char* argv[])
       if (projectType == "library")
         return "Static Library";
 
+      if (projectType == "dll")
+        return "Dynamic Library";
+
       if (projectType == "audioplug")
         return "Audio Plug-in";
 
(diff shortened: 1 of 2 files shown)


lethal-guitar:
On it!

Jucer2Reprojucer/main.cpp
```diff
@@ -908,7 +908,7 @@ int main(int argc, char* argv[])
           const auto configuration = configurations.getChild(i);
 
           wLn("jucer_export_target_configuration(");
-          wLn("  \"", std::get<1>(element), "\"");
+          wLn("  \"", element.second, "\"");
```
lethal-guitar: I assume this is coming from JUCE's XML library?

McMartin: No. `element` is a poorly named variable coming from this `for` loop:
https://github.com/McMartin/FRUT/blob/36354b6452b7eb0ae4b1be80ed7670a2aa5de2f3/Jucer2Reprojucer/main.cpp#L655-L665

I guess I should rename it to `exporterPair`, or simply use a `struct` instead of `std::pair`. But that's for another PR.

McMartin: Done in https://github.com/McMartin/FRUT/pull/300/commits/7c4c5955146da1de2e034c0ac5c0e08e77e00d4c.

PR #324, opened 2018-03-01
Build AUv3 plugins on 64-bit only

McMartin:


McMartin:
Projucer doesn't issue any warning, and AUv3 is 64-bit only by "definition", so I don't think Reprojucer should issue a warning.

PR #329, opened 2018-03-25
Prefer juce::String and juce::StringArray to std::string and std::vector<std::string>

McMartin:


MartyLake:
Lgtm!

PR #346, opened 2018-04-29
Convert the "Enable Inter-App Audio" audio plugin setting

McMartin:
This setting is only relevant for standalone audio plugins on iOS, but Reprojucer.cmake doesn't support building JUCE projects for iOS yet. This is why 51fb241 doesn't add `ENABLE_INTERAPP_AUDIO` to the docs and doesn't add a warning that mentions a GitHub issue.


MartyLake:
Lgtm ☺️

PR #349, opened 2018-05-24
Make it possible to use Cotire (precompiled headers) with FRUT,  et al

scotchi:
We're in the process of switching over to FRUT from our own internal CMake hackery.  These commits get things to the point that we can drop FRUT in to replace our internal system.

Unlike most projects, we're not moving from Projucer, so the first change is to remove that assumption from the system.  We ditched Projucer a year or two ago and removed the file from our repo.

Next is a simple fix to apply the `-fpic` flag to shared libraries on Linux.

The final change is much larger, and presumably more controversial.  One of the main reasons we switched to CMake was to be able to take advantage of precompiled headers, which speed up builds for JUCE-based projects dramatically.

However, if you try to build JUCE itself using a precompiled header that aggregates all JUCE headers, it fails with this error:

    #error "Incorrect use of JUCE cpp file"

The most sane way to work around that (which we already did in our internal CMake stuff) was to create a separate static library for JUCE itself, then generate a PCH for the JUCE library that's used when compiling the application code.

In our case, using Cotire (CMake framework for PCH) makes builds roughly twice as fast when doing a clean build, and even more when doing incremental builds of application code (since JUCE code rarely changes).

In our case, these are the build times for `make -j 6` on a 4 core system when adding Cotire to a JUCE-based project with about 150 source files:

With:

    real	1m25.629s
    user	3m17.779s
    sys	0m20.408s

Without:

    real	1m54.487s
    user	6m0.534s
    sys	0m31.057s

On system with less cores the "real" time is more dramatic since this introduces a couple of single-process bottlenecks (generating and compiling the PCH, building and linking the extra static lib).

I implemented this by adding a `CREATE_JUCE_LIBRARY_TARGET` option to `jucer_export_target()` (as best as I could tell, that seemed the most sane place to add such).  When that's on, a separate target is added, which can be used with Cotire.  With Cotire, we use a really simple self-generated header, basically:

    #ifdef __APPLE__
    #define Point CarbonPoint
    #include <MacTypes.h>
    #undef Point
    #endif

    #include "JuceLibraryCode/JuceHeader.h"

Right now I modified things with the assumption that deeper Cotire integration would be unwanted, but I can imagine that being useful itself and a built-in option itself.  If you'd prefer to have things that way, I'd be happy to rework the patches where it's just something like `PRECOMPILED_HEADERS_ENABLE ON`.

If that flag isn't set, the only side effect should be that now the static library is named `lib${PROJECT}_SharedCode` instead of `lib${PROJECT}`.  That didn't seem meaningful enough to work around.


McMartin:
@scotchi thank you so much for this Pull Request! I really enjoy the very comprehensive description. 

The three commits are very different, so they could have been submitted as separate Pull Requests, but I guess I should just blame the lack of contribution guidelines. 😅

cmake/Reprojucer.cmake
```diff
@@ -1860,12 +1862,6 @@ endfunction()
 
 function(_FRUT_abs_path_based_on_jucer_project_dir in_path out_path)
 
-  if(NOT IS_ABSOLUTE "${in_path}" AND NOT DEFINED JUCER_PROJECT_DIR)
-    message(FATAL_ERROR "The path \"${in_path}\" must be absolute, unless you give "
-      "PROJECT_FILE when calling jucer_project_begin()."
-    )
-  endif()
-
```
McMartin: `Reprojucer.cmake` is meant to be used by developers who are not familiar with CMake and this fatal error is very useful to catch beginner mistakes. We should not delete it. Instead, we should extend it to make it a bit more useful to beginner users, as well as advanced users. See https://github.com/McMartin/FRUT/pull/350.

cmake/Reprojucer.cmake
```diff
@@ -53,6 +53,8 @@ function(jucer_project_begin)
     endif()
     get_filename_component(project_dir "${_PROJECT_FILE}" DIRECTORY)
     set(JUCER_PROJECT_DIR "${project_dir}" PARENT_SCOPE)
+  else()
+    set(JUCER_PROJECT_DIR "${CMAKE_CURRENT_SOURCE_DIR}" PARENT_SCOPE)
```
McMartin: @scotchi since you don't have a `.jucer` file, you indeed don't have a `PROJECT_FILE` to give to `jucer_project_begin()`. However, there is nothing preventing your from defining `JUCER_PROJECT_DIR` manually in your CMakeLists.txt:
```cmake
set(JUCER_PROJECT_DIR "<project_dir>")
```
where `<project_dir>` is `${CMAKE_CURRENT_LIST_DIR}`, `${CMAKE_CURRENT_SOURCE_DIR}`, `${CMAKE_SOURCE_DIR}`, or something else, depending on the structure of your JUCE project. I added that option to the error message from `_FRUT_abs_path_based_on_jucer_project_dir()` in #350.

scotchi:
Is there any interest in the main bit of this (optional separate target for JUCE code) being merged eventually?  For us this was just the first in a handful of changes we wanted to mess around with, and I wasn't sure if we should look towards continuing to maintain our own fork (including merging in patches from master) or if something like this has a chance of making it in upstream.  We've also got one bugfix that went into our tree that's no longer in this pull request.

McMartin:
@scotchi there is concrete interest in supporting separate targets for JUCE modules (including user-defined modules). I'm currently, but slowly, working on this. I can @-mention you on Pull Requests that go into that direction, so you can share your opinion. Talking about your opinion, you would still like to get it on https://github.com/McMartin/FRUT/pull/351.

PR #350, opened 2018-05-24
Improve fatal error message about absolute path

McMartin:

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -1866,6 +1866,10 @@ function(_FRUT_abs_path_based_on_jucer_project_dir in_path out_path)
       "    jucer_project_begin(\n"
       "      PROJECT_FILE \"<path/to/YourProject.jucer>\"\n"
       "    )\n"
+      "If you don't have a .jucer file for your project, you can instead define "
+      "JUCER_PROJECT_DIR manually:\n"
+      "    jucer_project_begin()\n"
+      "    set(JUCER_PROJECT_DIR \"\${CMAKE_CURRENT_[LIST|SOURCE]_DIR}\")\n"
     )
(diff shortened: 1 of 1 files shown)


scotchi:
Seems fine -- but is there a particular reason *not* to assume that `${CMAKE_CURRENT_SOURCE_DIR}` is the project root?  If you're not building JUCE itself, are there common cases where you wouldn't want that?

McMartin:
I had to re-read the CMake documentation to refresh my understanding of `CMAKE_CURRENT_LIST_DIR` vs. `CMAKE_CURRENT_SOURCE_DIR` vs. `CMAKE_SOURCE_DIR`. While doing so I realized that depending on the project structure, someone might want `Reprojucer.cmake` to assume that the project root is any of them, or even something else.
Being explicit about the value of `JUCER_PROJECT_DIR` should prevent subtle bugs, especially when moving/merging/splitting CMakeLists.txt files.

scotchi:
I suppose I'm coming from a more CMake-y background (though I realize reading over my changes that I'm somewhat rusty there too).  To me felt natural to have `jucer_project_files(...sources...)` assume the same things as `add_library(...sources...)` about their paths.  (And in my case, that's exactly what I was porting things from.)

I don't actually care very much about this -- it just felt weird having to be explicit about the source root when using Reprojucer when you wouldn't normally have to with CMake.

McMartin:
> To me felt natural to have `jucer_project_files(...sources...)` assume the same things as `add_library(...sources...)` about their paths.

> it just felt weird having to be explicit about the source root when using Reprojucer when you wouldn't normally have to with CMake.

@scotchi these are very good points, and I agree with you. So I opened #351 and closing this PR in its favor.

PR #362, opened 2018-07-04
Premature close to foreach in jucer_export_target

scotchi:
The subject says most of it -- there's a block here that assumes that `${config}` is defined, but the `foreach()` that defines it has already been closed.  I wasn't sure if more smaller iterations would be preferable to one large block like this, but leave that as an exercise for the reader... ;-)


scotchi:
I should note:  all this commit does is move the `endforeach()` and then adjust the indentation.  `git diff -w` may be useful.

scotchi:
Oh, this is very weird.  This isn't the patch I originally sent.  I tried to rebase to master, and it seems git tried to be very clever and moved this to a completely different part of the file.

scotchi:
Going back to this, I have no idea how exactly I landed at this solution -- I suspect I got to trusting some of my large collection of debugging lines a bit too much, and was hurrying trying to get off to lunch.

#363 has the right fix.

PR #363, opened 2018-07-04
Variable shouldn't be evaluated before check for definition

scotchi:
`DEFINED` should be used on the variable name, not its value


McMartin:
@scotchi thank you so much for finding and fixing this bug!
Could you please reword the commit message to make it start with a verb? (like all the other commits of FRUT) Thanks!

scotchi:
I've updated the commit message now.

PR #365, opened 2018-08-08
Make JUCE module files browsable in the IDE or not per module

McMartin:


cmake/Reprojucer.cmake
```diff
@@ -457,15 +465,17 @@ function(jucer_project_module module_name PATH_KEYWORD modules_folder)
...
-  endforeach()
-  list(APPEND JUCER_PROJECT_BROWSABLE_FILES ${browsable_files})
-  set(JUCER_PROJECT_BROWSABLE_FILES ${JUCER_PROJECT_BROWSABLE_FILES} PARENT_SCOPE)
+  if(make_juce_code_browsable)
+    file(GLOB_RECURSE browsable_files "${modules_folder}/${module_name}/*")
+    foreach(file_path ${browsable_files})
+      get_filename_component(file_dir "${file_path}" DIRECTORY)
+      string(REPLACE "${modules_folder}" "" rel_file_dir "${file_dir}")
+      string(REPLACE "/" "\\" sub_group_name "${rel_file_dir}")
+      source_group("Juce Modules${sub_group_name}" FILES "${file_path}")
```
MartyLake: Shouldn't this be    `     source_group("Juce Modules ${sub_group_name}" FILES "${file_path}") `, with a [space] between Juce Modules and sub group name ?

McMartin: This line didn't change (it was only indented) in this PR. But the code is correct, `sub_group_name` will be something like `\juce_core\containers` or `\juce_events\messages`.

MartyLake:
You're right ! LGTM THEN :)

PR #377, opened 2018-08-25
Improve usage of set() command

McMartin:


MartyLake:
Lgtm!

Unisng unset everywhere instead of unset and set would be an easier code style IMHO

McMartin:
@MartyLake you are right, using `unset` everywhere is easier to apply when writing the code, but I find it more confusing when reading it. 

Consider the following two examples:
```cmake
    unset(header_search_paths)
    foreach(path ${_HEADER_SEARCH_PATHS})
      # ...
      list(APPEND header_search_paths "${path}")
    endforeach()
```
```cmake
    set(header_search_paths "")
    foreach(path ${_HEADER_SEARCH_PATHS})
      # ...
      list(APPEND header_search_paths "${path}")
    endforeach()
```
I find the first one more confusing, since it raises a few questions:
- what does it mean to `list(APPEND)` on an undefined variable?
- if `_HEADER_SEARCH_PATHS` is empty, the body of the `foreach` loop won't be executed and  `header_search_paths` will stay undefined, does this have an impact on the code that uses `header_search_paths` later on?

So in the end I will keep commit a7a19af.

MartyLake:
Yes, I get it. Then the code style will be : "Any variable that will be used later and defined internally should be defined, with a value or an empty string."

Issue #9, opened 2016-10-07
Add option to remove warnings from JUCE


MartyLake:
Atm, when using introjucer, compilers think juce source code is our code. And this library triggers various warnings. As a user, I don't want to adjust the warning level of my project depending on the warnings of the JUCE library, nor fix JUCE library to avoid warnings.

MartyLake:
There is the "SYSTEM" option for include directories https://cmake.org/cmake/help/v3.0/command/include_directories.html
But I could not find any similar option for add_executable.

PR #380, opened 2018-09-05
Support compiling JUCE modules in separate targets

McMartin:
This PR adds the `COMPILE_IN_TARGET` keyword to the `jucer_project_module()` command, which enables compiling JUCE module source files in specific targets.

Fixes https://github.com/McMartin/FRUT/issues/9.
Supersedes https://github.com/McMartin/FRUT/pull/349.


MartyLake:
This is a new feature compared to the original juce! I love it! Lgtm

scotchi:
I managed to get this working in my project, but it still misses setting a lot of essential flags.  Here's what I had to add to get it working (aside from the obvious instances of `COMPILE_IN_TARGET`):

    add_library(JuceLibrary STATIC)
    _FRUT_set_compiler_and_linker_settings(JuceLibrary)
    target_include_directories(JuceLibrary PRIVATE External/JUCE/modules)
    target_compile_definitions(JuceLibrary PRIVATE
      $<$<CONFIG:Debug>:DEBUG=1 _DEBUG=1> $<$<CONFIG:Release>:NDEBUG=1 _NDEBUG=1>)

    # ...

    set(shared_code ${PROJECT_NAME}_Shared_Code)
    target_link_libraries(${shared_code} PUBLIC JuceLibrary)

It feels like most of that should be taken care of automatically, but isn't.

McMartin:
The `jucer_project_end()` command takes care of creating the target you pass with `COMPILE_IN_TARGET`, and of setting the compiler settings and custom Xcode flags:
https://github.com/McMartin/FRUT/blob/dd95f096f4128c0c1657752d5a47f6a5777bad1f/cmake/Reprojucer.cmake#L1456-L1461

This means you shouldn't create the `JuceLibrary` target yourself. You can see that in action in `Jucer2Reprojucer/CMakeLists.txt`. The 3 JUCE modules that are used (`juce_core`, `juce_data_structures`, and `juce_events`) are compiled in a `JuceModules` target, but that target is not created beforehand. After `jucer_project_end()` has been called, the `JuceModules` target exists and its settings can be modified:
https://github.com/McMartin/FRUT/blob/2233b070de7909ce0ec7ecd4636dbd7fd19088dd/Jucer2Reprojucer/CMakeLists.txt#L201-L209

I'll add a warning in `jucer_project_end()` if the target already exist and I'll extend the documentation to make this clearer.

scotchi:
Ah, in our source we were referring to the target before `jucer_project_end()`, so it didn't exist, and I thought I had to create it.

Once working around that, we still run into issues with this line:

    add_library(${module_target} OBJECT ${module_sources})

Specifically, by being `OBJECT` (which otherwise makes sense) instead of `STATIC` it means that we can't easily use it from another helper target within the same project.  (We have an updater which also links to JUCE and for obvious reasons has to be a separate target.)  Just using _FRUT_set_compiler_and_linker_settings() on our helper target also didn't resolve things.

McMartin:
As you noted, `${module_target}` is an `OBJECT` target. Its source files are only compiled into object files. You can refer to these object files with the generator expression `$<TARGET_OBJECTS:${module_target}>`. 

In your case, it would look something like this:
```cmake
add_executable(Updater
  ${updater_sources}
  $<TARGET_OBJECTS:JuceLibrary>
)
```

However, your original idea of creating the target before is maybe easier to deal with for users of `Reprojucer.cmake`. I'll give it some thoughts and maybe create another PR with that approach...

scotchi:
Yes, I'd tried that, but then linking fails since the libraries required by JUCE aren't pushed out to the additional target via PUBLIC or PRIVATE.

McMartin:
Sorry for taking so long to understand what you are trying to do, but now I think I get it 😅 

Unfortunately, `Reprojucer.cmake` doesn't support compiling the JUCE modules once and then using them in several targets (like a plugin and a helper executable in your case, if I got it right). And I don't think `Reprojucer.cmake` will support that use case, because it is about reproducing what Projucer does.

However, FRUT is not just about reproducing what Projucer does, and I do plan to add another way of integrating JUCE and CMake, which will use the "pure" CMake way of doing so (`find_package` & cie.). I already started working on this several months ago (on the branch [`wip/add-files-for-find_package`](https://github.com/McMartin/FRUT/tree/wip/add-files-for-find_package)) and I actually plan to resume that work very soon.

@scotchi Maybe I should park this PR for now and focus on the `Find*.cmake` files instead. What do you think?

scotchi:
Yeah, sounds like it might work.  I'll try it a little later tonight.  As I was thinking about it, I also wonder if `COMPILE_AS_TARGET` or `CREATE_TARGET` might make more sense as a label.  `COMPILE_IN_TARGET` feels more like it'd imply to me that the target should already exist.

scotchi:
Just a little update on this -- I've been messing around with it off and on, but at the moment it seems to build all objects twice instead of reusing the objects from the other target.  I also haven't verified this, but based on build speed, it seems it's not using precompiled headers, so the build slows down by about 4x all told.  Again, I'll dig in more in the next day or two.

scotchi:
We had a release recently, so now I'm finally cycling back to build system cleanup and was hoping to eventually get us back on upstream FRUT.  There are still two issues:

- FRUT leaves all of its source variables defined, so a second run picks up on some of them.  Specifically `JUCER_PROJECT_FILES` is appended to, so any subsequent targets also include all sources of the previous ones.  This is what I described vaguely earlier as slowing down compilation so much (and since we weren't using PCH on the the smaller target, made it seem so slow).

I fixed this by adding:

```
function(jucer_project_begin)

  # Clear variables if this is not the first jucer project in this file.

  get_cmake_property(variables VARIABLES)
  foreach(variable ${variables})
    if("${variable}" MATCHES "JUCER_")
      message(STATUS "Clearing ${variable}")
      unset(${variable} PARENT_SCOPE)
    endif()
  endforeach()
  ...
```

I originally put that more sensibly in `jucer_project_end()`, but then realized that our project actually uses some of those `JUCER_*` variables for additional configuration.

- When reusing the JuceLibrary target (from our `COMPILE_AS_TARGET` param), it doesn't get linked into the second instance jucer project.  I'm still investigating this.

scotchi:
I did get this working in my tree based on your branch.  It needed the two commits I added here:

https://github.com/decomposer/FRUT/tree/support-compiling-modules-in-separate-targets

Specifically, in the code in your branch, it overwrites the OBJECT target created by the first run, which then fails in a handful of ways.

scotchi:
Also, thinking out loud, it made me wonder if it might make sense to *always* create object libraries for *each* JUCE module by default.  It'd need a sanity check to make sure the flags were the same, but that would obviate the need for the `COMPILE_AS_TARGET` flag completely (basically, if the same flags were used, it would reuse the OBJECT target by default.

That may be optimizing for too odd of a corner case, but I thought I'd at least mention the idea...

McMartin:
@scotchi Happy New Year! Thanks a lot for coming back to that PR. 

> FRUT leaves all of its source variables defined, so a second run picks up on some of them.

That is normally not an issue, since each project should be in a separate directory, and these variables are scoped to the current directory. This means that the workaround I proposed in https://github.com/McMartin/FRUT/pull/380#issuecomment-419043235 is wrong, and the two projects (`Banana` and `Banana-updater`) should be defined in separate directories, i.e. in separate CMakeLists.txt files.

> I fixed this by adding:

Clearing the `JUCER_*` variables in `jucer_project_begin` is a no-go, since users are encouraged to define some `JUCER_*` variables (for instance `JUCER_VST3_SDK_FOLDER`) on the command line.

> Also, thinking out loud, it made me wonder if it might make sense to *always* create object libraries for *each* JUCE module by default.

This might lead to object libraries being created but not used (because there are several jucer projects that use different flags), leading to extra compilation, which is what we're trying to avoid in the first place.

In the end, I think this PR doesn't deliver a good solution to being able to build JUCE modules only once and then use them in several jucer projects. I rebased it on top of the latest `master`, but I won't work more on it. Instead, I am focusing on writing `Find*.cmake` files, which should be a better alternative to `Reprojucer.cmake` for you and other developers who want to build JUCE projects "the CMake way".

scotchi:
Ok.  I've reworked our project to where we mostly can live without this feature.  Specifically, I have Cotire working (which I'll do a little write-up on at some point, since it speeds up compilation a lot, and I now have it working without changes to FRUT).  I have to recompile juce_core twice, but that's not the end of the world (that's all our helper needs), and that gets us back on master, which is I think good for everyone.  (Dealing with bugs from a forked version is a PITA.)

One thing I will note though, is that the JUCER_* variables absolutely do have to be cleared to create a second `jucer_project` even from a subdirectory.  I assume this is because the relevant values (most notably `JUCER_PROJECT_FILES`) are created with `PARENT_SCOPE`.  At present, to get things working, I have to use my clearing code at the top of our subdirectory project.

McMartin:
> even from a subdirectory

Sorry, I should have been more specific and I should have said "sibling directories" instead of "separate directories". 

From https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variables:
> Each of the Directories in a source tree has its own variable bindings. Before processing the `CMakeLists.txt` file for a directory, CMake copies all variable bindings currently defined in the parent directory, if any, to initialize the new directory scope.

So instead of
```cmake
jucer_project_begin()
  ...
jucer_project_end()

add_subdirectory(helper)
```
you need to do
```cmake
add_subdirectory(helper)

jucer_project_begin()
  ...
jucer_project_end()
```

McMartin:
Closing this PR since it doesn't provide an acceptable solution.

cor3ntin:
Are you still working on a solution in this design space?
It would be very helpful!

McMartin:
Hi @cor3ntin,
Thank you for your interest in FRUT!
I am not working on this topic, but this solution was not great anyway.
Could you please [create a new issue](https://github.com/McMartin/FRUT/issues/new) to explain what you are trying to achieve and what problems you are facing when using FRUT? I would more than happy to help you!

PR #382, opened 2018-09-12
Support the "Plugin VST3 Category" project setting

### LINKED ISSUE #375: Reprojucer.cmake doesn't support the setting PLUGIN_VST3_CATEGORY
It would be great for `Reprojucer.cmake` to support the setting `PLUGIN_VST3_CATEGORY` ("Plugin VST3 Category" in Projucer).

As of Projucer 5.3.1, this is also a multiple choice value.

Expected output in `AppConfig.h`:
```cpp
#ifndef  JucePlugin_Vst3Category
 #define JucePlugin_Vst3Category           "Fx|Delay"
#endif
```

McMartin:
Resolves #375.

@rclement: FYI


McMartin:
@MartyLake following our discussion via PM, I decided to split the second commit into 3 commits. This should make the review really easier since the commits of this PR now match closely the commits of JUCE. Please have another look, thanks!

PR #383, opened 2018-09-17
Support the "Plugin Formats" and "Plugin Characteristics" project settings

McMartin:

### DIFF
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -498,19 +498,52 @@ int main(int argc, char* argv[])
     if (projectType == "audioplug")
     {
       wLn("jucer_audio_plugin_settings(");
-      convertOnOffSettingWithDefault(jucerProject, "buildVST", "BUILD_VST", true);
-      convertOnOffSettingWithDefault(jucerProject, "buildVST3", "BUILD_VST3", false);
-      convertOnOffSettingWithDefault(jucerProject, "buildAU", "BUILD_AUDIOUNIT", true);
-      convertOnOffSettingWithDefault(jucerProject, "buildAUv3", "BUILD_AUDIOUNIT_V3",
-                                     false);
-      convertOnOffSettingWithDefault(jucerProject, "buildRTAS", "BUILD_RTAS", false);
-      convertOnOffSettingWithDefault(jucerProject, "buildAAX", "BUILD_AAX", false);
-      if (jucerVersionAsTuple >= Version{5, 0, 0})
+
+      if (jucerVersionAsTuple >= Version{5, 3, 1})
       {
-        convertOnOffSettingWithDefault(jucerProject, "buildStandalone",
-                                       "BUILD_STANDALONE_PLUGIN", false);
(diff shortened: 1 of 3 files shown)


Jucer2Reprojucer/main.cpp
```diff
@@ -498,19 +498,52 @@ int main(int argc, char* argv[])
...
+          {
+            if (vars.contains(allVars[i]))
+            {
+              values.add(allValues[i]);
+            }
+          }
+          return values;
+        };
+
+        const juce::StringArray formatsVars = {
```
MartyLake: This is pretty difficult to read. Why not put them in the lambda, or name the lamba variables like the callsite ?

MartyLake: s/allVars/formatVars/
s/allValues/formatValues/

MartyLake: This would be more straighforward if written as a translation list:

```cpp
const auto convertVarsToValues =
    [](const juce::var& v, const std::vector<std::pair<juce::String, juce::String>>& varToValues) -> juce::StringArray {
    auto vars = juce::StringArray::fromTokens(v.toString(), ",", {});
    juce::StringArray values;
    for (const auto& [ var, value ] : varToValues)
        if(vars.contains(var)) {
            values.add(value);              
       }
    }
    return values;
};
const auto varToValues = {
    {"buildVST", "VST"},  
    {"buildVST3", "VST3"},
// etc.
};
convertSettingAsList(jucerProject, "pluginFormats", "PLUGIN_FORMATS",
                             [&](const juce::var& v) {
                               return convertVarsToValues(v, varToValues);
                             });
```

McMartin:
@MartyLake based on your great feedback (thank you so much!), I improved the first two commits:
- I renamed the lambda from `convertVarsToValues` to `convertIdsToStrings`
- I made it take a map instead of two arrays
- I inlined the static data into the other lambdas, removing the need for confusing names

Please have another look, thank!

Jucer2Reprojucer/main.cpp
```diff
@@ -498,19 +498,52 @@ int main(int argc, char* argv[])
...
       {
-        convertOnOffSettingWithDefault(jucerProject, "buildStandalone",
-                                       "BUILD_STANDALONE_PLUGIN", false);
-        convertOnOffSettingWithDefault(jucerProject, "enableIAA", "ENABLE_INTERAPP_AUDIO",
+        const auto convertIdsToStrings =
+          [](const juce::var& v,
+             const std::map<juce::String, const char*>& idsToStrings) {
+            const auto ids = juce::StringArray::fromTokens(v.toString(), ",", {});
+            juce::StringArray strings;
+            for (const auto& idToString : idsToStrings)
```
MartyLake: replace with cpp17 structured binding for even better readability ? (fromhttps://skebanga.github.io/structured-bindings/ )
```cpp
for (const auto&& [ id, string ] : idsToStrings)
{
    if (ids.contains(id))
              {
                strings.add(string);
              }
}
```

MartyLake: This version is way better than the 1st proposal. maybe a cpp17 feature is not a necessary constraint for FRUT, therefore I'm accepting the changes. feel free to include or not this suggestion.

PR #391, opened 2018-09-27
Support adding entitlements to AUv3 plugins

### LINKED ISSUE #373: .entitlements issue
Its seems so that there are successful builds with cmake, but I could not find anything to set the sandbox flag that is under the .entitlements in a project.
This means that the project can not be run inside from GarageBand. 
Also I could not find anything on .nib extensions

McMartin:
Resolves #373.


McMartin:
@MartyLake ping

PR #406, opened 2018-10-20
Add Console Application test projects created with Projucer

McMartin:
This PR adds several test projects that will be used in a subsequent PR. Nothing very interesting here...

@MartyLake: please "review" (have a look at where the files are added and whether the commit messages make sense, but don't waste time looking at the contents of the files), thanks!


McMartin:
@MartyLake ping

PR #407, opened 2018-10-20
Speed-up CI by re-using the same binary directory

McMartin:


McMartin:
CMake doesn't like when you change the generator, or when you change CACHEd variables. But here `JUCE_ROOT`, `CMAKE_BUILD_TYPE` and `JUCE_VERSION` are not CACHE variables, so we're good.

PR #423, opened 2018-11-14
Support converting modules that use global paths

McMartin:

Big thanks to @adishavit for https://github.com/adishavit/argh! 👏


MartyLake:
LGTM! kuddos to @adishavit for https://github.com/adishavit/argh!  it's pleasing to see a library that is readable, simple and efficiently relying on the STD <3

PR #427, opened 2018-11-17
Fix passing a "-m" compiler flag when giving "<None>" to ARCHITECTURE

### LINKED ISSUE #426: Release builds set to -march=native on Linux, making them incompatible with many machines
https://github.com/McMartin/FRUT/blob/282f7204f3bada25ea411e209180a4fa2384de20/cmake/Reprojucer.cmake#L1358

It seems almost certainly wrong to use `-march=native` on release builds since it stands to reason that portable binaries are desired.   Before realizing that FRUT was adding this flag, I spent several hours tracking down why VSTs built on my Linux desktop with this were crashing on startup when I installed the packages on my laptop.  Specifically, the default x86-64 should probably not be overridden.

McMartin:
Closes #426.


MartyLake:
LGTM

PR #447, opened 2018-12-03
Treat VST plugin format as legacy like Projucer does

McMartin:


McMartin:
The files under `ci/fake-SDKs` are only here to avoid unnecessary warnings on CI when configuring all JUCE projects. Adding more files would be more confusing than anything else.

PR #456, opened 2019-01-13
Fix non-Xcode macOS version targeting

scotchi:
Currently neither `OSX_DEPLOYMENT_TARGET` nor `OSX_BASE_SDK_VERSION` work for non-Xcode builds.  This fixes two issues to get things working for both of them.  This can be verified with the trivial script:

```
#!/bin/bash

otool -l $1 | grep -A4 LC_VERSION_MIN_MACOSX | head -4
```

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -3618,7 +3618,7 @@ function(_FRUT_set_compiler_and_linker_settings target)
           OUTPUT_STRIP_TRAILING_WHITESPACE
         )
         if(IS_DIRECTORY "${sysroot}")
-          target_compile_options(${target} PRIVATE "-isysroot ${sysroot}")
+          target_compile_options(${target} PRIVATE -isysroot "${sysroot}")
         else()
           message(WARNING "Running `xcrun --sdk macosx${sdkroot} --show-sdk-path` didn't"
             " output a valid directory."
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -3609,6 +3609,9 @@ function(_FRUT_set_compiler_and_linker_settings target)
       target_compile_options(${target} PRIVATE
         "-mmacosx-version-min=${osx_deployment_target}"
       )
+      set_property(TARGET ${target} APPEND_STRING PROPERTY LINK_FLAGS
+        " -mmacosx-version-min=${osx_deployment_target}"
+      )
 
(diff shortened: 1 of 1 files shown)


cmake/Reprojucer.cmake
```diff
@@ -3618,7 +3618,7 @@ function(_FRUT_set_compiler_and_linker_settings target)
           OUTPUT_STRIP_TRAILING_WHITESPACE
         )
         if(IS_DIRECTORY "${sysroot}")
-          target_compile_options(${target} PRIVATE "-isysroot ${sysroot}")
+          target_compile_options(${target} PRIVATE -isysroot ${sysroot})
```
McMartin: @scotchi it is advised to quote paths, in case they contain spaces, or worse semicolons. Please keep double-quotes around `${sysroot}`.

cmake/Reprojucer.cmake
```diff
@@ -3609,6 +3609,9 @@ function(_FRUT_set_compiler_and_linker_settings target)
       target_compile_options(${target} PRIVATE
         "-mmacosx-version-min=${osx_deployment_target}"
       )
+      target_link_options(${target} PRIVATE
+        "-mmacosx-version-min=${osx_deployment_target}"
+      )
```
McMartin: `target_link_options` was recently introduced in CMake 3.13, but the minimum CMake version required for FRUT is 3.4. Please replace with
```cmake
      set_property(TARGET ${target} APPEND PROPERTY LINK_FLAGS
        "-mmacosx-version-min=${osx_deployment_target}"
      )
```

scotchi:
Updated.

scotchi:
Sorry for the previous push; I should have looked more closely before pushing.  The rebase went awry.

This one *doesn't* put quotes around the path in the output (but neither would your suggestion) -- not sure if you were suggesting that was needed?  Since the path is generated by the script itself based on input that's already been through validity checking, it's not entirely clear to me why that's needed.

cmake/Reprojucer.cmake
```diff
@@ -3609,6 +3609,9 @@ function(_FRUT_set_compiler_and_linker_settings target)
       target_compile_options(${target} PRIVATE
         "-mmacosx-version-min=${osx_deployment_target}"
       )
+      set_property(TARGET ${target} APPEND_STRING PROPERTY LINK_FLAGS
```
McMartin: Good job on figuring out that `LINK_FLAGS` is a string, not a list. This means we need to fix other places where we are appending to `LINK_FLAGS` as if it were a list. But I can do that on another PR.

PR #461, opened 2019-02-03
Find macOS frameworks from the right SDK

McMartin:


McMartin:
@MartyLake: thanks for the quick review, but I'm actually not satisfied with the last commit. I want to refine the behavior, but that will have to wait until I can get my hands on a Mac. I'll open a new PRs with the other commits though.

PR #492, opened 2019-05-05
Fix missing linker flag when enabling "Link-Time Optimisation"

### LINKED ISSUE #491: /LTCG linker option does not turned on in Visual Studio when building a library.
This happens either when I'm using a default Link-Time Optimisation, or when I enable it explicitly.
Can you check this out?
Thank you!

McMartin:
Resolves #491.

@czyjerry: FYI


MartyLake:
Lgtm! :D

czyjerry:
Thanks for the fix!
This fix the issue in all the stanalone app projects, and dynamic library projects.
But I found that this issue still exists in static library projects.
Can you check it again?

Thanks!

PR #502, opened 2019-05-10
Refactor generating .plist files

McMartin:
@MartyLake: please review, thanks!

PR #503, opened 2019-05-19
Add test projects for .plist files generation

McMartin:
I guess I should have added these test projects _before_ doing #502 (_Refactor generating .plist files_), but better late than never :smile:


McMartin:
@MartyLake ping

PR #508, opened 2019-08-01
Install AAX to home Library

brettbuddin:
Other plugin types' installation steps install to the user's `Library` directory. AAX installation isn't behaving like other plugin formats. It isn't until reading back through the logs that I realized that its installing to a different location from those formats. The first time you run a build after adding AAX support results in its inability to write to `/Library` due to permissions issues.

Plugin development seems fraught with strange behaviors so if this is an erroneous PR, let me know. It could be that this was done on purpose—that the user's `Library` will not load properly in ProTools. If this is the case, please shut this PR down.

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -2173,7 +2173,7 @@ function(jucer_project_end)
         endforeach()
 
         _FRUT_install_to_plugin_binary_location(${aax_target} "AAX"
-          "/Library/Application Support/Avid/Audio/Plug-Ins"
+          "$ENV{HOME}/Library/Application Support/Avid/Audio/Plug-Ins"
         )
       elseif(MSVC)
         set_property(TARGET ${aax_target} PROPERTY SUFFIX ".aaxdll")


McMartin:
Thanks a lot for your interest in FRUT and the PR!

Have you tried building and installing an AAX plugin using Projucer? You will have the same issue, since Projucer sets the "AAX Binary Location" to `/Library/Application Support/Avid/Audio/Plug-Ins/` by default (see https://github.com/WeAreROLI/JUCE/blob/dc1cf29cae85992f282af5d6519ccdaf73b7f6af/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_Xcode.h#L540).

`Reprojucer.cmake` tries to reproduce what Projucer does, especially wrt. defaults, since you can always override them by specifying the value that you want.

This is how it would look in your `CMakeLists.txt` file:
```cmake
  "Xcode (MacOSX)"
  NAME "Debug"
  DEBUG_MODE ON
  AAX_BINARY_LOCATION "$ENV{HOME}/Library/Application Support/Avid/Audio/Plug-Ins"
)

  "Xcode (MacOSX)"
  NAME "Release"
  DEBUG_MODE OFF
  AAX_BINARY_LOCATION "$ENV{HOME}/Library/Application Support/Avid/Audio/Plug-Ins"
)
```

brettbuddin:
Oh! Thank you for pointing the configuration option out. I agree that changing the default would be a bad idea if that’s what Projucer has. Closing this.

PR #509, opened 2019-09-07
Change juce::ScopedPointer to std::unique_ptr

franklange:
Change the use of juce::ScopedPointer to std::unique_ptr in order to adapt to the API change introduced in JUCE 5.4.4.

### DIFF
--- a/cmake/BinaryDataBuilder/extras/Projucer/Source/Utility/jucer_FileHelpers.cpp
+++ b/cmake/BinaryDataBuilder/extras/Projucer/Source/Utility/jucer_FileHelpers.cpp
@@ -47,6 +47,7 @@
 
 #include "../jucer_Headers.h"
 
+#include <memory>
 
 //==============================================================================
 namespace FileHelpers
@@ -85,7 +86,7 @@ namespace FileHelpers
 
     int64 calculateFileHashCode (const File& file)
     {
-        ScopedPointer<FileInputStream> stream (file.createInputStream());
+        std::unique_ptr<FileInputStream> stream{file.createInputStream()};
         return stream != nullptr ? calculateStreamHashCode (*stream) : 0;
     }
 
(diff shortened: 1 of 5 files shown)


McMartin:
Hi @franklange 

Thank you very much for this Pull Request! I was pretty busy this summer, so I didn't see that `juce::ScopedPointer` had been replaced by `std::unique_ptr` in so many places in JUCE.

As the state of CI shows (all red), we can't merge your changes as-is. The main issue is that FRUT should be compatible with any version of JUCE since 4.2.1 (when `Projucer` was introduced), so the FRUT code needs to support both `juce::ScopedPointer` and `std::unique_ptr`. If you know how to do that and want to do it, feel free to push new commits to this PR. The CI systems will tell you what works or doesn't. If you want me to take care of it, please open a GitHub issue (https://github.com/McMartin/FRUT/issues/new) to report the problem so I can credit you in the README (https://github.com/McMartin/FRUT#contributors) for "Bug reports". Then I'll make a PR to make FRUT support JUCE 5.4.4 (without dropping support for older JUCE versions).

franklange:
Hey,
thanks for the quick response! I saw that the builds were failing due to the old JUCE version but I was kinda waiting for a reply to discuss what I can do about it.

I've been thinking about what you wrote and I'm not sure if it's a good idea to keep backwards compatibility on this one. This is a breaking change in the API you're trying to wrap, so I think it makes sense to go with the flow.
Have you considered using tags for indicating the JUCE version? This way your master would always work with the latest JUCE version and if someone wants to use an older JUCE version, they can checkout the tagged release commit for that version...
This seems way more reasonable to me than trying to cover all JUCE versions with essentially the same code base...

Just for context: a friend of mine recently tried using your project but since it didn't even compile with the latest JUCE version immediately abandoned ship. I already had a local branch with the fix, so I knew what to do when he told me but that's kinda the worst case scenario.

Anyhoo, let me know how I can help. If you do want to continue trying to cover different return types with the same code base I'd rather opt-out and just create the bug ticket, since this doesn't really resonate with me. Other than that let me know what I can do!

franklange:
Hey,
thanks for the feedback! I opened an issue for this :)

Issue #520, opened 2019-09-18
Certain configuration names cause CMake errors


TheSlowGrowth:
I converted an existing application to a CMakeLists.txt file using `Jucer2Reproducer`. The original application contained exporter configurations with these names:

````
Debug (32bit)
Debug (64bit)
Release (32bit)
Release (64bit)
````

The generated CMakeLists.txt thus contains lines like these:
````
jucer_export_target_configuration(
   "Visual Studio 2017"
   NAME "Debug (32bit)"
   DEBUG_MODE ON
   # ...
)
````

When running cmake I get these errors:
````
CMake Error at FRUT/prefix/FRUT/cmake/Reproducer.cmake:4550 (target_link_libraries):
   Error evaluating generator expression:
      $<CONFIG:Debug (32bit)>
   Expression syntax not recognized.
````

If I remove the spaces and brackets so that the names are `Debug32bit`. `Release64bit`, etc. the error disappears. It appears to me that `${config}` in `$<CONFIG:${config}>` must be escaped in some way but according to the CMake documentation, only angular brackets, commas and semicolons have proper escape sequences.

I would love to learn how to escape these properly. If that's not possible it would be a good solution to convert the incompatible configuration names to something that is compatible when converting the project with `Jucer2Reproducer`. 

I would be happy to contribute a fix for this, but at this point I'm unsure which option to choose.

McMartin:
I found out that since [its introduction in 2012](https://gitlab.kitware.com/cmake/cmake/commit/9d9f616792ee07a460af9f0a6dc036d81b852e66#750111181780c34a5649e3ff3bc1150c13010784_28_30) up until [the latest release of CMake](https://gitlab.kitware.com/cmake/cmake/blob/v3.15.3/Source/cmGeneratorExpressionNode.cxx#L886), the `$<CONFIG:cfg>` generator expression requires `cfg` to match the regex `^[A-Za-z0-9_]*$`. This means that escaping characters that don't match won't work, and that we need to convert configuration names.

I'll work on a fix ASAP.

TheSlowGrowth:
Okay, that is weird. The doc clearly states that `>`, `,` and `;` can be escaped in a generator expression. Apparently `$<CONFIG:cfg>` has more strict requirements which are not documented.

I'll try to work a fix into `Jucer2Reproducer` if that's okay. This should be fairly easy to do.

McMartin:
Currently being worked on by @TheSlowGrowth in #521.

PR #521, opened 2019-09-19
Add configuration name sanitization to Jucer2Reproducer

TheSlowGrowth:
`Jucer2Reproducer` now removes invalid characters from configuration names and ensures that they are unique to each exporter by adding digits if required. If the name was changed, a warning is printed to cerr.

### DIFF
--- a/Jucer2Reprojucer/CMakeLists.txt
+++ b/Jucer2Reprojucer/CMakeLists.txt
@@ -103,7 +103,7 @@ jucer_project_module(
 
 jucer_export_target(
   "Xcode (MacOSX)"
-  COMPILER_FLAGS_FOR_maximum_warnings "-Werror -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded"
+  COMPILER_FLAGS_FOR_maximum_warnings "-Werror -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-exit-time-destructors -Wno-padded"
 )
 
 jucer_export_target_configuration(
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -2006,9 +2006,19 @@ int main(int argc, char* argv[])
       {
         const auto configuration = configurations.getChild(i);
 
+        const auto originalConfigName = configuration.getProperty("name").toString();
+        const auto sanitizedConfigName = sanitizeConfigurationName(originalConfigName);
+
         wLn("jucer_export_target_configuration(");
         wLn("  \"", exporterName, "\"");
-        wLn("  NAME \"", configuration.getProperty("name").toString(), "\"");
+        if (originalConfigName != sanitizedConfigName)
+        {
+          const auto jucerFileBasename = jucerFile.getFileName();
+          wLn("  NAME \"", sanitizedConfigName, "\" # \"", originalConfigName, "\" in ",
+              jucerFileBasename);
+        }
+        else
+          wLn("  NAME \"", sanitizedConfigName, "\"");
 
         const auto isDebug = bool{configuration.getProperty("isDebug")};
         wLn("  DEBUG_MODE ", (isDebug ? "ON" : "OFF"));
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -151,6 +151,82 @@ void printError(const juce::String& error)
 }
 
 
+/** Sanitizes a string to match the regex ^[A-Za-z0-9_]*$. Sanitization follows these
+ *  rules:
+ *  1) Invalid characters are replaced with "_".
+ *  2) At most one consecutive replacement "_" is kept (if the original name contained
+ *     multiple "_" they remain unaltered).
+ *  3) No replacement "_" is added at the end or at the beginning.
+ */
+juce::String sanitizeName(const juce::String& input)
+{
+  const juce::String validCharacters =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_";
+  juce::String output;
+  bool mustInsertReplacementChar = false;
+  for (auto c = input.getCharPointer(); !c.isEmpty();)
+  {
+    // keep valid characters
+    if (validCharacters.containsChar(*c))
+    {
+      if (mustInsertReplacementChar)
+      {
+        output += '_';
+        mustInsertReplacementChar = false;
+      }
+      output += *c;
+    }
+    // replace invalid characters with "_"
+    else
+    {
+      // don't add "_" at the beginning
+      if (output.isNotEmpty())
+        mustInsertReplacementChar = true;
+    }
+    c++;
+  }
+  return output;
+}
+
+/** Sanitizes a configuration name. Valid config names match the regex ^[A-Za-z0-9_]*$.
+ *  Sanitization is done with sanitizeName(). An optional digit is added until the
+ *  configuration name is unique. The sanitization is synchronized across
+ *  the entire project.
+ */
+juce::String sanitizeConfigurationName(const juce::String& originalName)
+{
+  static std::map<juce::String, juce::String> sanitizedNameTable;
+  static juce::StringArray sanitizedNameList;
+  juce::String sanitizedName;
+
+  // check if this name was sanitized before
+  if (sanitizedNameTable.find(originalName) != sanitizedNameTable.end())
+    return sanitizedNameTable[originalName];
+
+  // need to sanitize
+  sanitizedName = sanitizeName(originalName);
+
+  // add digits until unique
+  int number = 1;
+  auto sanitizedNameWithoutNumber = sanitizedName;
+  while (sanitizedNameList.contains(sanitizedName))
+    sanitizedName = sanitizedNameWithoutNumber + "_" + juce::String(number++);
+
+  // remember it for later
+  sanitizedNameTable[originalName] = sanitizedName;
+  sanitizedNameList.add(sanitizedName);
+
+  if (originalName != sanitizedName)
+  {
+    std::cerr << "warning: Configuration '" << originalName << "' "
+              << "contains invalid characters. The name was changed to '" << sanitizedName
+              << "'" << std::endl;
+  }
+
+  return sanitizedName;
+}
+
+
 juce::String makeValidIdentifier(juce::String s)
 {
   if (s.isEmpty())
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -691,14 +691,7 @@ int main(int argc, char* argv[])
     wLn();
   }
 
-  auto escapedJucerFileName = jucerFileName.toStdString();
-  std::replace_if(
-    escapedJucerFileName.begin(), escapedJucerFileName.end(),
-    [](const std::string::value_type& c) {
-      return !(std::isalpha(c, std::locale::classic())
-               || std::isdigit(c, std::locale::classic()));
-    },
-    '_');
+  auto escapedJucerFileName = sanitizeName(jucerFileName);
   const auto jucerFileCMakeVar = juce::String{escapedJucerFileName} + "_FILE";
 
   // get_filename_component() or set(*_FILE)


Jucer2Reprojucer/main.cpp
```diff
@@ -270,6 +270,51 @@ void writeUserNotes(LineWriter& wLn, const juce::ValueTree& valueTree)
...
+            }
+            else if (originalName != sanitizedName)
+                std::cerr << "warning: Configuration '" << originalName << "' in exporter '" << exporterName
+                          << "' contains invalid characters. The name was changed to '" << sanitizedName << "'" << std::endl;
+
+            // add digits until unique
+            int number = 1;
+            auto sanitizedNameWithOptionalNumber = sanitizedName;
+            while (sanitizedConfigNames.contains(sanitizedNameWithOptionalNumber))
+                sanitizedNameWithOptionalNumber = sanitizedName + juce::String(number++);
```
McMartin: This is a great idea! However, there is an edge case that this implementation doesn't cover: when exporters don't have the configurations in the same order, or when some configurations only exist on certain exporters.

For instance:
```yaml
- XCODE_MAC:
  - configurations:
    - name: "Config (-feature)"
      defines: "FEATURE=0"
    - name: "Config (+feature)"
      defines: "FEATURE=1"

- LINUX_MAKE:
  - configurations:
    - name: "Config (+feature)"
      defines: "FEATURE=1"
    - name: "Config (-feature)"
      defines: "FEATURE=0"
```
would be converted to 
```yaml
- XCODE_MAC:
  - configurations:
    - name: "Configfeature"
      defines: "FEATURE=0"
    - name: "Configfeature1"
      defines: "FEATURE=1"

- LINUX_MAKE:
  - configurations:
    - name: "Configfeature"
      defines: "FEATURE=1"
    - name: "Configfeature1"
      defines: "FEATURE=0"
```
which would break the matching of the configuration names between the exporters.

To solve this issue, we should collect all configuration names, make a mapping of original to sanitized configuration names, and then use that mapping when writing the configuration names to the `CMakeLists.txt` file.

Jucer2Reprojucer/main.cpp
```diff
@@ -270,6 +270,51 @@ void writeUserNotes(LineWriter& wLn, const juce::ValueTree& valueTree)
...
+
+        auto exporter = exporterSettingsWithInvalidConfigNames.createCopy();
+        juce::StringArray sanitizedConfigNames;
+
+        for (int i = 0; i < exporter.getNumChildren(); i++)
+        {
+            // remove invalid characters
+            auto configuration = exporter.getChild(i);
+            auto originalName = configuration.getProperty("name").toString();
+            auto sanitizedName = originalName.retainCharacters(validCharacters);
```
McMartin: Removing invalid characters does the job, but it's not very nice for configuration names like `Debug (all warnings)` or `Release (with LTO)`. 

Do you think you could come up with an algorithm that:
- replace invalid characters with `_`,
- keep at most one consecutive replacement `_` (i.e. it's fine to have several consecutive `_` characters if they were already in the original configuration name),
- don't add a replacement `_` at the end?

That would make `Debug (all warnings)` -> `Debug_all_warnings` and `Release (with LTO)` -> `Release_with_LTO`

Jucer2Reprojucer/main.cpp
```diff
@@ -270,6 +270,51 @@ void writeUserNotes(LineWriter& wLn, const juce::ValueTree& valueTree)
   }
 }
 
+    juce::ValueTree sanitizeConfigurationNames(const juce::String& exporterName, const juce::ValueTree& exporterSettingsWithInvalidConfigNames)
```
McMartin: I know it is not written down explicitly, and there is no CI check for it, but the C++ code of FRUT is formatted using `clang-format`. There is a [`.clang-format` ](https://github.com/McMartin/FRUT/blob/c3dd1775bba0ad0db43389203658151e4db77198/.clang-format) file at the root of the repository, which is currently configured to be used with `clang-format-7.0.0` (you can download binaries for most platforms on http://releases.llvm.org/download.html#7.0.0). I usually run the following:
```
clang-format-7.0.0 -i -style=file -verbose Jucer2Reprojucer/main.cpp
```

Jucer2Reprojucer/main.cpp
TheSlowGrowth: Sure, good idea

Jucer2Reprojucer/main.cpp
TheSlowGrowth: Hmm, I see. Yes, the matching would break. You're right, that should be avoided. I'll make the required changes.

Jucer2Reprojucer/main.cpp
TheSlowGrowth: `makeValidIdentifier` seems to do most of this already, except than only some characters are replaced with `_`, other are simply removed (this includes the space).

Now that we're dealing with multiple name-sanitization functions, it would be good to a) merge them into one or b) alter their names to be more clear about their use.
Forthe sake of readability and simplicity I would prefer to use `makeValidIdentifier` for the config name sanitization as well and add the required changes to this function instead of writing another one.

Do you agree?

Jucer2Reprojucer/main.cpp
TheSlowGrowth: Yes, sure I can run clang-format on it. I simply didn't notice it. Thanks for pointing it out.

I would suggest to add a check to the CI. [run-clang-format](https://github.com/Sarcasm/run-clang-format) could help with this.

Jucer2Reprojucer/main.cpp
McMartin: Jucer2Reprojucer's `makeValidIdentifier` is a stripped-down version of [a function named the same in Projucer](https://github.com/WeAreROLI/JUCE/blob/5.4.4/extras/Projucer/Source/Utility/Helpers/jucer_CodeHelpers.cpp#L71-L123). It is then used in Jucer2Reprojucer exactly like in Projucer. We shouldn't change the logic of this function, otherwise Jucer2Reprojucer will diverge from Projucer.  I agree that there should be some documentation (or at least a comment) on the function to explain that.

We could indeed use `makeValidIdentifier` to sanitize configuration names, but it wouldn't be as nice as the algorithm I proposed (`Debug (all warnings)` would become `Debugallwarnings`). 

If you are concerned about code reuse, don't worry, this new function to convert a string into a (nice looking) valid CMake identifier will be perfect to compute `escapedJucerFileName` as well:
https://github.com/McMartin/FRUT/blob/dfd36f97e0c654f379d3f7a6034d50be2efef326/Jucer2Reprojucer/main.cpp#L546-L552

Jucer2Reprojucer/main.cpp
McMartin: > I would suggest to add a check to the CI.

I'll work on that when I have some time 😅

McMartin: Done in https://github.com/McMartin/FRUT/pull/526.

TheSlowGrowth:
@McMartin Does this look good to you? Still some cleanup and squashing left to do but the direction should hopefully be correct. I'll continue tomorrow.
Side note: clang-format messes with the comments. Could comment reflowing be switched off? Otherweise I'll exclude the comments from clang-format, if requried.

TheSlowGrowth:
@McMartin please review

Jucer2Reprojucer/main.cpp
```diff
@@ -270,6 +270,106 @@ void writeUserNotes(LineWriter& wLn, const juce::ValueTree& valueTree)
...
+}
+
+// clang-format off
+/** Reads a jucerProject and sanitizes configuration names. Valid config names match the
+  * regex ^[A-Za-z0-9_]*$. Sanitization is done with sanitizeName(). An optional digit is 
+  * added until the configuration name is unique. The sanitization is synchronized across 
+  * all exporters in the project.
+  */
+// clang-format on
+void sanitizeConfigurationNames(const juce::ValueTree& jucerProject)
```
McMartin: This function is pretty well written, but it doesn't fit the architecture of `Jucer2Reprojucer`. More concretely, `Jucer2Reprojucer`:
- never modifies the original `ValueTree`,
- extracts data from the `ValueTree` right before transforming it and writing it to the `CMakeLists.txt` file, and
- tries to read the content of the `ValueTree` only once.

Could you please try to take the content of this function and move it to where we are already looping over `EXPORTFORMATS` and `CONFIGURATIONS`? This would guarantee that we are iterating in the same way when sanitizing and when outputting to the `CMakeLists.txt` file. It would also allow us to write the original configuration name as a comment when the sanitized name is different:
```cmake
  "Visual Studio 2017"
  NAME "Debug_32_bit"  # "Debug (32-bit)" in MyGreatProject.jucer
  DEBUG_MODE ON
)
```
Finally, the warnings would be properly interleaved with other warnings that we might add in the future.

Jucer2Reprojucer/main.cpp
```diff
@@ -543,14 +645,7 @@ int main(int argc, char* argv[])
...
 
-  auto escapedJucerFileName = jucerFileName.toStdString();
-  std::replace_if(
-    escapedJucerFileName.begin(), escapedJucerFileName.end(),
-    [](const std::string::value_type& c) {
-      return !(std::isalpha(c, std::locale::classic())
-               || std::isdigit(c, std::locale::classic()));
-    },
-    '_');
+  auto escapedJucerFileName = sanitizeName(jucerFileName);
```
McMartin: Please make this cleanup in its own commit. I think there should be 3 commits in this PR:
- one for introducing the `sanitizeName` function,
- one for using the `sanitizeName` function on configuration names, and
- one for using the `sanitizeName` function here.

McMartin:
@TheSlowGrowth Happy new year! Thanks a lot for all the good work on this PR. I would like to see it merged in the coming month. Do you intend to continue working on it or can I take over?

McMartin:
@TheSlowGrowth I would like to see this PR merged and #520 closed in a week or two. Do you have time to work on this topic or should I handle it myself? If I don't hear from you by February 8th, 2020, I'll  rebase. rework and merge this PR on my own.

TheSlowGrowth:
@McMartin I wasn't logged into github for quite some time (RL ...) and just saw your messages. Please excuse that I didn't see this earlier.
I have just pushed a new attempt. Please take a look if this fixes all issues. I have tested it myself and it seems fine, though I'm afraid I might have missed some spots where the sanitization would have to be added.

McMartin:
@TheSlowGrowth no problem at all, you're entirely excused.

I'll do a quick review now and see how we should handle clang. Maybe silencing new warnings is fine.

Jucer2Reprojucer/main.cpp
```diff
@@ -151,6 +151,81 @@ void printError(const juce::String& error)
...
+  if (sanitizedConfigNameTable.find(originalName) != sanitizedConfigNameTable.end())
+    return sanitizedConfigNameTable[originalName];
+
+  // need to sanitize
+  sanitizedName = sanitizeName(originalName);
+
+  // add digits until unique
+  int number = 1;
+  auto sanitizedNameWithoutNumber = sanitizedName;
+  while (sanitizedConfigNameTable.find(sanitizedName) != sanitizedConfigNameTable.end())
```
McMartin: `sanitizedConfigNameTable.find(sanitizedName)` will always return `sanitizedConfigNameTable.end()` because `sanitizedName` is only used as a value of the map, not as a key.
Using 2 `juce::StringArray` objects, one for the original names and one for the sanitized names, might result in simpler code.

Jucer2Reprojucer/main.cpp
```diff
@@ -151,6 +151,82 @@ void printError(const juce::String& error)
...
+      output += *c;
+    }
+    // replace invalid characters with "_"
+    else
+    {
+      // don't add "_" at the beginning
+      if (output.isNotEmpty())
+        mustInsertReplacementChar = true;
+    }
+    c++;
```
McMartin: This happens on every iteration of the loop, so it should be in the third clause of the `for` statement.

Jucer2Reprojucer/main.cpp
```diff
@@ -151,6 +151,82 @@ void printError(const juce::String& error)
...
+ */
+juce::String sanitizeConfigurationName(const juce::String& originalName)
+{
+  static std::map<juce::String, juce::String> sanitizedNameTable;
+  static juce::StringArray sanitizedNameList;
+  juce::String sanitizedName;
+
+  // check if this name was sanitized before
+  if (sanitizedNameTable.find(originalName) != sanitizedNameTable.end())
+    return sanitizedNameTable[originalName];
```
McMartin: If `originalName` was sanitized before, the function will return here without having used `sanitizedName` at all. Please declare and initialize `sanitizedName` only when you need it.

Jucer2Reprojucer/main.cpp
```diff
@@ -151,6 +151,82 @@ void printError(const juce::String& error)
...
+
+/** Sanitizes a configuration name. Valid config names match the regex ^[A-Za-z0-9_]*$.
+ *  Sanitization is done with sanitizeName(). An optional digit is added until the
+ *  configuration name is unique. The sanitization is synchronized across
+ *  the entire project.
+ */
+juce::String sanitizeConfigurationName(const juce::String& originalName)
+{
+  static std::map<juce::String, juce::String> sanitizedNameTable;
+  static juce::StringArray sanitizedNameList;
```
McMartin: Similarly to `sanitizedName`, please declare and initialize `sanitizedNameList` only when you need it. Even if it is a `static` variable, it doesn't need to be at the top of the function.

Jucer2Reprojucer/main.cpp
```diff
@@ -2006,9 +2006,19 @@ int main(int argc, char* argv[])
...
         wLn("  \"", exporterName, "\"");
-        wLn("  NAME \"", configuration.getProperty("name").toString(), "\"");
+        if (originalConfigName != sanitizedConfigName)
+        {
+          const auto jucerFileBasename = jucerFile.getFileName();
+          wLn("  NAME \"", sanitizedConfigName, "\" # \"", originalConfigName, "\" in ",
+              jucerFileBasename);
+        }
+        else
+          wLn("  NAME \"", sanitizedConfigName, "\"");
```
McMartin: Please put curly braces around this statement.

Jucer2Reprojucer/CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ jucer_project_module(
 
 jucer_export_target(
   "Xcode (MacOSX)"
-  COMPILER_FLAGS_FOR_maximum_warnings "-Werror -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded"
+  COMPILER_FLAGS_FOR_maximum_warnings "-Werror -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-exit-time-destructors -Wno-padded"
```
McMartin: Adding `-Wno-exit-time-destructors` to the compiler flags of `Jucer2Reprojucer` should be done in the same commit that introduces the static variables, otherwise that commit won't build.

McMartin:
Superseded by https://github.com/McMartin/FRUT/pull/655.

PR #523, opened 2019-09-23
Support precompiled libraries inside JUCE modules

### LINKED ISSUE #517: Static libraries in JUCE modules don't end up in the linker flags
I have a JUCE module `myJuceModule` that contains static libraries (according to the JUCE module header documentation). I have a project `myProject` that uses this module. It works just fine when using the Projucer. 

When I migrate  `myProject` with `Jucer2Reprojucer` the static libraries from `myJuceModule` don't end up in the "Other Linker Flags" setting in XCode. Include paths are fine and the module is actually build. But obviously it fails at the linker stange where the symbols from the static libraries are missing.

McMartin:
Resolves #517.

@TheSlowGrowth: please test that it works for you, thanks!


TheSlowGrowth:
Thank you so much! This is great!

I tried it out. Here's what I did:
1. Pulled branch `support-precompiled-libraries-in-modules`
2. Copied it to `<myProjectRoot>/FRUT`
3. Followed the `Getting Started` instructions in Readme.rst
4. Tried to generate project files for VS 2017 with cmake

Output was:
````
Configuring done
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_EXE_LINKER_FLAGS_DEBUG32BIT
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_EXE_LINKER_FLAGS_RELEASE32BIT
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_EXE_LINKER_FLAGS_DEBUG64BIT
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_EXE_LINKER_FLAGS_RELEASE64BIT
Generating done
````

Not sure if it's an error on my side, though.

Also some questions:
Firstly, I have never seen `target_link_libraries` with `-LIBPATH` or `-L` arguments. Is that official syntax? I wonder if this can guarantee that libraries contained in the module will be chosen over system libraries with the same name.

Also, I wonder if it would be possible to allow for finer granularity in the OSX lib paths. On Windows it is possible to have precompiled libraries for different versions of Visual Studio as well as Debug and Release versions in the JUCE module (simply because all these properties appear in the search path). On OSX, there is basically just one path - making it impossible to work with multiple versions of XCode and forcing me to link against release versions of the libraries even in debug builds. Maybe it would be possible to follow some logic like this: 
````
libSearchPath = "${moduleDir}/libs/${XCodeVersion}/${arch}/${debug/release}/";
if not exists(libSearchPath)
   libSearchPath = "${moduleDir}/libs/${arch}";
```` 
This would stay compatible to the Projucer definition but also allow to work with multiple XCode versions as well as debug/release builds of the precompiledd libraries. Honestly, I don't understand why the JUCE Module Definition didn't allow this in the first place.
I know this is asking for a lot and maybe this is the wrong place for this discussion. Just wanted to leave it here and hear your thoughts on it.

Thanks again for your work!

McMartin:
> Output was:
> 
> ```
> Configuring done
> CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
> Missing variable is:
> CMAKE_EXE_LINKER_FLAGS_DEBUG32BIT
> CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
> Missing variable is:
> CMAKE_EXE_LINKER_FLAGS_RELEASE32BIT
> CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
> Missing variable is:
> CMAKE_EXE_LINKER_FLAGS_DEBUG64BIT
> CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
> Missing variable is:
> CMAKE_EXE_LINKER_FLAGS_RELEASE64BIT
> Generating done
> ```
> 
> Not sure if it's an error on my side, though.

These errors are related to #520. When using a "Visual Studio" generator, CMake expects some `*_<CONFIG>` variables to exist. Maybe we should fix #520 first and then come back here?

---

> Firstly, I have never seen `target_link_libraries` with `-LIBPATH` or `-L` arguments. Is that official syntax?

From the [documentation of `target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html):
> This command has several signatures as detailed in subsections below. All of them have the general form
> ```
> target_link_libraries(<target> ... <item>... ...)
> ```
> ...
> Each `<item>` may be:
> ...
> - **A link flag**: Item names starting with `-`, but not `-l` or `-framework`, are treated as linker flags.
> ...

This is the best way I know of to specify where the linker should search for libraries.

The command [`target_link_directories`](https://cmake.org/cmake/help/latest/command/target_link_directories.html) was added in [CMake 3.13](https://cmake.org/cmake/help/latest/release/3.13.html#commands) but `Reprojucer.cmake` currently needs to work with CMake 3.4, so we can't use it.

---
> Also, I wonder if it would be possible to allow for finer granularity in the OSX lib paths.

I don't mind adding this extra feature to `Reprojucer.cmake`, but let's keep the scope of this PR and #517 focused on supporting what is defined in the JUCE Module Format. Feel free to create another issue (which would depend on #517) to propose this extra feature, so we can discuss it separately.

TheSlowGrowth:
> These errors are related to #520

I expected #520, so I sanitized the config names manually after I had generated the file with the current release version of `Jucer2Reproducer`. To my knowledge, this error should not occur with all config names only containing valid characters. Can you point me to where this error could come from? I wasn't able to find the spot where the variable should have been set and it seems multiple native cmake commands could be relevant here.

> let's keep the scope of this PR and #517 focused on supporting what is defined in the JUCE Module Format

I agree. I'll open another issue for this.

McMartin:
> Can you point me to where this error could come from?

Sorry for not explaining in more details in my previous reply. It's quite a long story, but here we go:

When using a Visual Studio generator, CMake sets [`CMAKE_CONFIGURATION_TYPES` ](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html) to `Debug;Release;MinSizeRel;RelWithDebInfo` by default. CMake also sets some all-caps variables that correspond to these configurations:
- `CMAKE_CXX_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_C_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_EXE_LINKER_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_MODULE_LINKER_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_RC_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_SHARED_LINKER_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`
- `CMAKE_STATIC_LINKER_FLAGS_{DEBUG,MINSIZEREL,RELEASE,RELWITHDEBINFO}`

You can find all these variables in the `CMakeCache.txt` file. Here is for instance the `CMakeCache.txt` generated by CMake when configuring an empty `CMakeLists.txt` with the `Visual Studio 14 2015` generator: [CMakeCache.txt](https://github.com/McMartin/FRUT/files/3649327/CMakeCache.txt)

In `Reprojucer.cmake`, we set `CMAKE_CONFIGURATION_TYPES` to the configuration names passed to `jucer_export_target_configuration`:
https://github.com/McMartin/FRUT/blob/dfd36f97e0c654f379d3f7a6034d50be2efef326/cmake/Reprojucer.cmake#L1665
but we don't set any of the above-mentioned variables for these configurations. So when you call `jucer_export_target_configuration` with `Debug32bit`, `Release32bit`, `Debug64bit`, and `Release64bit`, `CMAKE_CONFIGURATION_TYPES` is set to `Debug32bit;Release32bit;Debug64bit;Release64bit`, but `CMAKE_EXE_LINKER_FLAGS_{DEBUG32BIT,RELEASE32BIT,DEBUG64BIT,RELEASE64BIT}` are not set.

When CMake generates the Visual Studio project files, it wants to append [`CMAKE_EXE_LINKER_FLAGS_<CONFIG>`](https://cmake.org/cmake/help/latest/variable/CMAKE_EXE_LINKER_FLAGS_CONFIG.html) to [`CMAKE_EXE_LINKER_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_EXE_LINKER_FLAGS.html) for each build configuration in `CMAKE_CONFIGURATION_TYPES` and it complains if `CMAKE_EXE_LINKER_FLAGS_<CONFIG>` is not defined (see https://gitlab.kitware.com/cmake/cmake/blob/v3.15.3/Source%2FcmLocalVisualStudio7Generator.cxx#L937-941 for the actual code).

The solution to this issue is to make `Reprojucer.cmake` define all the `CMAKE_*_FLAGS_<CONFIG>` variables appropriately (either to empty, or to `CMAKE_*_FLAGS_DEBUG` or `CMAKE_*_FLAGS_RELEASE` based on whether `DEBUG_MODE ON` was passed to `jucer_export_target_configuration`). This is something I would like to fix as part of fixing #520, since it is related to build configurations.

McMartin:
@TheSlowGrowth: I've rebased this PR on top of the latest `master`, which now includes #527, so you should be able to test it without any CMake error about missing variables. Please let me know if you can build your console application that uses a custom JUCE module with precompiled libraries. Thanks!

@MartyLake: please wait for @TheSlowGrowth to report back before reviewing, thanks!

TheSlowGrowth:
Hello!
Thank you very much for your effort. I finally found time to try it and can confirm that it works for our project (I tried it with Visual Studio).
However, in cmake-gui the various variables now appear as settings and flood the whole screen (see attached screenshot). This makes it hard to see the actual configuration options. Could this be resolved?

FRUT_cmake_gui

McMartin:
> However, in cmake-gui the various variables now appear as settings and flood the whole screen (see attached screenshot). This makes it hard to see the actual configuration options. Could this be resolved?

I'm sorry. This is a bug, and I've opened https://github.com/McMartin/FRUT/pull/530 to fix it.

McMartin:
@MartyLake: this PR is now ready for review, thanks!

McMartin:
> last commit is very concise compared to the feature it implements

Glad you like it!

PR #527, opened 2019-09-29
Support more than just "Debug" and "Release" as build configurations

McMartin:

@TheSlowGrowth: FYI. This fixes the error messages from CMake about "missing variable" that you reported in https://github.com/McMartin/FRUT/pull/523#issuecomment-534429155.


MartyLake:
LGTM!

PR #528, opened 2019-09-30
Handle configuration names that are not valid for CMake in Reprojucer.cmake

McMartin:
This is resolving #520 partially. What is left is making Jucer2Reprojucer write valid configuration names when converting `.jucer` files, and is being worked on in #521 by @TheSlowGrowth.


MartyLake:
LGTM!

PR #536, opened 2019-10-08
Print the output from try_compile() when it fails

McMartin:


McMartin:
I'm closing this now, since I want to rework the process for reporting issues with building and installing BinaryDataBuilder, IconBuilder, and PListMerger.

Issue #513, opened 2019-09-15
Failed to build and install BinaryDataBuilder


McMartin:
When `Reprojucer.cmake` fails to build and install `BinaryDataBuilder`, it prints the following error message:

```
-- Building and installing BinaryDataBuilder
CMake Error at <root>/FRUT/cmake/Reprojucer.cmake:2992 (message):
  Failed to build and install BinaryDataBuilder.  Please report this problem
  by writing a new comment on the following GitHub issue:
  https://github.com/McMartin/FRUT/issues/513
```

If you are getting this error message when configuring one of your projects, please write a new comment at the end of this issue to report it. You can write a comment even if the issue is closed, however it won't re-open it automatically.

DustVoice:
-- Setting CMAKE_BUILD_TYPE to "Debug" as it was not specified.
-- Building and installing BinaryDataBuilder
CMake Error at D:/FRUT/cmake/Reprojucer.cmake:3199 (message):
  Failed to build and install BinaryDataBuilder.  Please report this problem
  by writing a new comment on the following GitHub issue:
  https://github.com/McMartin/FRUT/issues/513
Call Stack (most recent call first):
  D:/FRUT/cmake/Reprojucer.cmake:1906 (_FRUT_generate_JuceHeader_header)
  CMakeLists.txt:316 (jucer_project_end)


-- Configuring incomplete, errors occurred!

McMartin:
Hi @DustVoice,

Thanks for trying out FRUT!

Could you please configure this project again from scratch (i.e. without any files generated by CMake around) and reply with the exact command that you ran and the complete output from `cmake`?
Here is an example:
```
$ cmake .. -DJUCE_ROOT=../../JUCE
-- Building for: Visual Studio 14 2015
-- The C compiler identification is MSVC 19.0.24215.1
-- The CXX compiler identification is MSVC 19.0.24215.1
-- Check for working C compiler using: Visual Studio 14 2015
-- Check for working C compiler using: Visual Studio 14 2015 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler using: Visual Studio 14 2015
-- Check for working CXX compiler using: Visual Studio 14 2015 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
[...]

PR #541, opened 2019-10-10
Add the "big object files" as options for MinGW

DustVoice:
Pass the `-mbig-obj` option to the assembler, in case of using MinGW on windows, in order to prevent an error regarding too big object files, which is a known error for MinGW. This fixes it!

(As discussed in https://github.com/McMartin/FRUT/issues/513)


McMartin:
Great commit message! Thanks a lot.

PR #542, opened 2019-10-19
Upgrade IconBuilder

McMartin:
Quite similar to what was done on `BinaryDataBuilder` in #413.

### DIFF
--- a/cmake/IconBuilder/main.cpp
+++ b/cmake/IconBuilder/main.cpp
@@ -69,7 +69,7 @@ int main(int argc, char* argv[])
       const auto iconFile = outputDir.getChildFile("Icon.icns");
 
       MemoryOutputStream outStream;
-      projectExporter.writeIcnsFile(images, outStream);
+      projectExporter.writeIcnsFile_v4_2_0(images, outStream);
 
       if (!FileHelpers::overwriteFileWithNewDataIfDifferent(iconFile, outStream))
       {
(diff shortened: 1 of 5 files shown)


cmake/IconBuilder/main.cpp
```diff
@@ -69,7 +98,14 @@ int main(int argc, char* argv[])
       const auto iconFile = outputDir.getChildFile("Icon.icns");
 
       MemoryOutputStream outStream;
-      projectExporter.writeIcnsFile_v4_2_0(images, outStream);
+      if (jucerVersion < Version{5, 4, 0})
```
MartyLake: NICE!!!!!

cmake/IconBuilder/main.cpp
```diff
@@ -39,10 +40,38 @@ int main(int argc, char* argv[])
...
-  const auto& iconFormat = args.at(1);
-  const auto outputDir = File{args.at(2)};
-  const auto smallIconImageFile = args.at(3) == "<None>" ? File{} : File{args.at(3)};
-  const auto largeIconImageFile = args.at(4) == "<None>" ? File{} : File{args.at(4)};
+  using Version = std::tuple<int, int, int>;
+
+  const auto jucerVersion = [&args]() {
+    if (args.at(1) == "latest")
+    {
+      return Version{5, 4, 0};
```
MartyLake: Should we extract "latest" to the top of this file ?

McMartin: I don't think that's valuable. When reading the code which compares `jucerVersion` to decide which function to call, you wouldn't know what `latest` is, until you jump to the top of the file. I prefer to keep that knowledge localized, even if it's a bit duplicated.

McMartin:
@MartyLake: thanks for the review! FYI, I pushed 2 commits with a few includes that I had forgotten. Feel free to have a look, but since they are pretty trivial, you don't have to review them.

Issue #473, opened 2019-03-19
Default architecture is i386, while through the Projucer it isn't


stijnfrishert:
When using FRUT to generate an XCode project through my generated CMakeLists.txt, Xcode throws up the following error:

    error: The i386 architecture is deprecated. You should update your ARCHS build setting to remove the i386 architecture.

When I generate an XCode project from my Projucer directly, the architecture is fine.

The OSX Architecture field in my Projucer is set to Default. When I change it to 64-bit Intel, the generated XCode project works, so it looks like the default isn't being detected correctly by FRUT.

McMartin:
@stijnfrishert thanks for trying out FRUT and creating that issue!

Projucer sets a lot of Xcode settings, and I admit that I didn't make `Reprojucer.cmake` set all of them in the exact same way. However, I think I found the Xcode setting that `Reprojucer.cmake` doesn't set that is responsible for that warning. I pushed a branch called [`wip/set-ONLY_ACTIVE_ARCH-Xcode-setting`](https://github.com/McMartin/FRUT/tree/wip/set-ONLY_ACTIVE_ARCH-Xcode-setting) with a potential fix. Please give it a try, and let me know if that solves your issue, thanks!

stijnfrishert:
Yes, that works! Thanks for the quick reaction. :)

Project still doesn't fully build to missing frameworks, but that's another issue. Architecture works as expected now.

McMartin:
As explained in https://github.com/McMartin/FRUT/pull/474#issuecomment-475041586, the changes I did on the branch `wip/set-ONLY_ACTIVE_ARCH-Xcode-setting` were not correct. So let's take a step back and try to understand what is really going on:
- which version of macOS, Xcode and CMake are you using?
- what kind of project are you building? 
  - if you can, please attach the `CMakeLists.txt` generated by `Jucer2Reprojucer` in a comment (feel free to remove/change any line that contains "sensitive" information)
  - if you can, please also attach the `.jucer` file that you are asking `Jucer2Reprojucer` to convert (again, feel free to obfuscate its contents) 
- where exactly do you see the error that you reported in your first post?
  - if in the Xcode UI, in which panel? When doing which action exactly?
  - if in the terminal, when running which command?

stijnfrishert:
Ok, I'll rerun this when I find the time! Just upgraded that system to Mojave yesterday actually, so the situation might've changed too.

stijnfrishert:
[...]

PR #543, opened 2019-10-29
Avoid error about deprecated i386 architecture

McMartin:
Resolves https://github.com/McMartin/FRUT/issues/473.

@stijnfrishert: I was finally able to reproduce the issue when building the PushNotificationsDemo example project from JUCE 5.2.1 with Xcode 10.1 on macOS 10.13 High Sierra. Sorry for the very long delay.


MartyLake:
LGTM

Issue #458, opened 2019-01-16
iOS and Android Support


scisci:
Hi, I got here from this JUCE forum post:

https://forum.juce.com/t/introducing-juce-cmake-build-your-juce-projects-using-cmake/22717

Its quite old so I'm just wondering, are iOS and Android still not supported? If not what is the challenge?

McMartin:
Hi,

Thank you for your interest in FRUT! 

The README lists the [supported exporter targets from Projucer](https://github.com/McMartin/FRUT/tree/a3d2b590bfb94e079c644a5f1a8f729e50018b68#supported-export-targets), and indeed FRUT still doesn't support "Xcode (iOS)" nor "Android".

For iOS, I simply can't do the work alone. I don't own a computer that runs macOS and I don't own an iPhone or iPad. Also, there are some settings that Projucer writes in the Xcode project (for instance [`SystemCapabilities`](https://github.com/WeAreROLI/JUCE/blob/baf78b1e5f6e02ade4f209d7f45a60e43f9b53cd/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_Xcode.h#L959-L964)), which CMake doesn't support. However, it doesn't mean that it's impossible to add some iOS support to FRUT. I have enough experience with the "Xcode (MacOSX)" exporter and enough knowledge about the Projucer code that I could implement some experimental iOS support "blindly". I just need to be sure that someone will be keen on testing my broken implementation and will report the issues.

For Android, it would be a bit simpler, since Android Studio is available on Linux and Windows, and I do own a smartphone that runs Android. However, I don't have any experience building for mobile, so it will require a fair amount of work from my side (especially considering all the changes on the "Android" exporter in Projucer since JUCE 4.2.0). Again, it is not impossible to do. All I need to get started are people like you who are showing their interest and will actually use what I've worked on.

If you're interested in helping me by testing some experimental support of the "Xcode (iOS)" and "Android" exporters in FRUT, let me know!

scisci:
Hi @McMartin that makes sense. if I do end up going with JUCE for an upcoming project I will help out. 

One other thing...I haven't actually tried your cmake code, but if I read it right it seems to need a Projucer file to work off. My hope was that a CMake JUCE build could do away with the need to run Projucer at all. I realize some decisions need to be made like whether its going to be an AudioUnit or something similar, but couldn't those just be CMAKE variables or flags of some kind?

[...]

PR #551, opened 2019-11-06
Support the "Xcode (iOS)" Projucer exporter

McMartin:
Related to #458.

This PR is made of the following commit "groups":
- commit 7fe3eafe29e464699327d779c3bdd2a9d7e1f113 documents per-exporter CMake requirements
- commits 6828d9b3ef3ad97f86d0c0ad22a805a00a03ab95 and 940cf8b1036b2b82b07d344493bf0a63684257bf only deal with renaming
- commits from e9fb724dcb7c6419e06f28df710eb3d5f0e7044c to a742a5def03ad56acc667bb4b36b4781868e7917 add the core of the iOS support
- commits from 887ade9ea5fec783753e0f23de695766a5fd31f6 to ee2314e241467496e312f03f9ea4b419d672c120 update generated `CMakeLists.txt` files (most of the changes were done by `Jucer2Reprojucer`)
- commits from 50919a512ce99f91cedaadb7eb2fe038be505c62 to 4e181866521bc1e5cb3058e4de84fa6c64128f26 add support for various settings that are specific to the "Xcode (iOS)" exporter
- commits 9e786c3a1bd6af7709450e9cd31a189681de5b0e and dc6bc364b37732d898492c401b9da29ea334d8d0 extend CI and polish the README


cmake/XcassetsBuilder/main.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  const std::vector<std::string> args{argv, argv + argc};
+
+  const auto xcassetsBundle = File::getCurrentWorkingDirectory()
+                                .getChildFile(juce::String{args.at(1)})
+                                .getChildFile("Images.xcassets");
+  const auto smallIconImageFile =
+    args.at(2) == "<None>"
+      ? File{}
+      : File::getCurrentWorkingDirectory().getChildFile(juce::String{args.at(2)});
```
MartyLake: Will this also work if path is absolute ?

cmake/Reprojucer.cmake
```diff
@@ -1938,6 +1988,12 @@ function(jucer_project_end)
     )
   endif()
 
+  if(JUCER_IOS)
+    _FRUT_create_xcassets_folder_from_icons(JUCER_XCASSETS)
+
```
MartyLake: empty line leftover ?

MartyLake:
LGTM, even if it's hard to say I read every single line of it

cmake/XcassetsBuilder/main.cpp
McMartin: It will indeed work. From https://github.com/WeAreROLI/JUCE/blob/5.4.5/modules/juce_core/files/juce_File.h#L243-L257:
```cpp
        If the string is actually an absolute path, it will be treated as such, e.g.
            File ("/moose/fish").getChildFile ("/foo.txt") will produce "/foo.txt"
```

cmake/Reprojucer.cmake
McMartin: Addressed in https://github.com/McMartin/FRUT/pull/551/commits/74dc8d7e4588683dad9cc1c60d642d14d51f2d94#diff-d7ac1cfde9bf2d7ef9cccf6fab14bc6eR1990-R1991.

McMartin:
[First-class support for iOS was added in CMake 3.14](https://cmake.org/cmake/help/latest/release/3.14.html#platforms). When I opened this PR, I didn't want to make that a hard requirement, since the code of FRUT itself only requires CMake version 3.4 minimum. However, after a bit of testing on macOS, and considering that Visual Studio 2019 is also only supported since CMake 3.14, I changed my mind. In this new version of the PR, CMake 3.14 is required for iOS. Instead of calling `cmake -G Xcode -DJUCER_IOS=ON` and using `JUCER_IOS` in `Reprojucer.cmake`, we now call `cmake -G Xcode -DCMAKE_SYSTEM_NAME=iOS` and we use [`IOS`](https://cmake.org/cmake/help/v3.14/variable/IOS.html) in `Reprojucer.cmake`.

@MartyLake: please re-review this PR. To help you understand what changed, you can read the [diff between bb0cb70 and 96933d1](https://github.com/McMartin/FRUT/compare/bb0cb703b8ac1782718991cd38b64ccc0401fc1b..96933d1f15cb2fd787c9d36d4e79c4ce5c20360d). I also updated the PR description with the new commit groups. Thanks!

McMartin:
We hit a new problem. `Reprojucer.cmake` needs to build its helper executables (`BinaryDataBuilder`, `IconBuilder`, `PListMerger`, and the new `XcassetsBuilder`), but these executables can only be built and run on macOS, not iOS. Since `Reprojucer.cmake` uses the `try_compile` CMake command, `CMAKE_SYSTEM_NAME=iOS` (and everything that comes with it) is passed down when configuring the helper executables. This obviously fails because the Cocoa framework cannot be found in the `iphoneos` SDK.

@MartyLake: please don't review yet, fixing this problem might require quite some changes.

McMartin:
After reading more about `try_compile` I came to the conclusion that it is not the right tool in the first place. `try_compile` should be used to determine if something can/would compile, it shouldn't be used to just build and install other CMake projects at configure time. I'll open a new PR to replace `try_compile` by something else in `_FRUT_build_and_install_helper_exe`, and then I'll rebase this one again. Stay tuned!

McMartin:
@MartyLake: I've rebased the branch on top of the latest `master` after merging #553 and force-pushed. This PR is now ready for review again, please have another look, thanks!

MartyLake:
LGTM, no obvious typo 🗡

PR #552, opened 2019-11-06
Fix double installation of bundle resources

McMartin:
This PR fixes a regression that was introduced in #544 and which is blocking #551.


MartyLake:
Lgtm ;D

PR #553, opened 2019-11-08
Support building and installing helper executables when cross-compiling

McMartin:
This is required to support iOS (see https://github.com/McMartin/FRUT/pull/551#issuecomment-551646978).


MartyLake:
LGTM

PR #554, opened 2019-11-11
Call install() per configuration

McMartin:
This cleanup/refactoring was found when working on the iOS support.


MartyLake:
YUP, much cleaner to just lean on the existing cmake tools instead of creating a complex expression :) LGTM

PR #558, opened 2019-11-17
add docker support

sabjorn:
Adds `Dockerfile` for creating a Docker image which contains the ``Jucer2Reprojucer`` and a helper script (`entrypoint.sh`) for easy integration with command line.

Docker image is based on `debian:stretch-light` and is 57MB in size.

Additionally: updated with instructions for building and running docker image.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,6 +110,31 @@ Now we can build ``MyGreatProject`` using CMake: ::
 including Ninja, NMake Makefiles (on Windows), Unix Makefiles (on Linux and macOS), Visual
 Studio 2013, 2015 and 2017 (on Windows), and Xcode (on macOS).
 
+Docker
+------
+
+Build
+^^^^^
+To build the Docker image, run ::
+
+  $ cd <root>/FRUT/docker
+  $ docker build -t frut:latest .
+
+This resulting image will contain ``Jucer2Reprojucer`` and ``Reprojucer.cmake`` as well as an ``entrypoint.sh`` for convenince 
+
+Run
+^^^
+To convert a `.jucer` file, run ::
+
+  $ docker run --rm -v <juce_project_directory>:/work frut /work/<juce_project>.jucer
+
+The ``.cmake`` file will be created in the same directory as the ``.jucer`` file. 
+
+Additionally, the Docker image can be entered for manual operation ::
+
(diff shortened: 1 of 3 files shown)


McMartin:
Hi @sabjorn,

I'm sorry for the very late reply. I was quite busy preparing for ADC'19 in London, then attending it, and then working for a week in Berlin. Now I'm back home and I have finally some time to take a proper look at your pull request.

I like the idea of using Docker to simplify the access to FRUT, but there are a few things that we need to consider before committing to a solution:
1. copying `Reprojucer.cmake` to the root of the Docker container means that the generated `CMakeLists.txt` will contain something like:
   ```cmake
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}//")
   ```
   which most likely doesn't match where `Reprojucer.cmake` actually is on the user machine.
2. in the (near) future, FRUT will also have a `FindJUCE.cmake` module and an associated `Jucer2CMake` executable, but we can't have 2 entrypoints on the Docker container AFAIK. How would we deal with that then?
3. would you still need a Docker container if there were binary releases of FRUT?

Let me know what's your take on these. 

To solve 1., I'm working on making `Reprojucer.cmake` optional when calling `Jucer2Reprojucer`. That way `Reprojucer.cmake` won't even have to be copied in the Docker container.

sabjorn:
Thanks for getting back to me. Glad to start this conversation!:

1. yes, this will certainly be a problem. Short term solution would be to copy the `Reprojucer.cmake` into the directory with the generated `CMakeLists.txt`. A little dirty since as a user I might be confused. Your solution sounds good though.
2. the entrypoint in the Docker image can be any script. This mean we can create a command line esq interface when launching the container. So any alternate runtime functionality can be enabled that way.
3. Binary release would reduce the need for a Docker container for sure. Or, more specifically, doing a binary release would mean supporting Docker directly in repo would be unnecessary.

It was rad to meet you at ADC! Hope I can be of use with this project!

McMartin:
> Or, more specifically, doing a binary release would mean supporting Docker directly in repo would be unnecessary.

Then I'll work on making a binary release as soon as I'm done with `FindJUCE.cmake`.

sabjorn:
makes sense to me. i'll close this PR. I'll keep an eye out for the binary release! Thanks!!!

PR #567, opened 2019-12-25
Fix converting SDK paths based on plugin formats

### LINKED ISSUE #566: FRUT doesn't consistently find `buildXXX` settings in recent JUCE 5 jucer files
Looks like only some of the code was updated to handle recent JUCE 5 jucer files which puts the `buildXXX` settings under `pluginFormats` rather than under the root properties of the project.

This causes, for example, building the AAX format with the resulting converted `CMakeLists.txt` file to fail, because the AAX SDK folder never gets written into the file, because the following condition is `false`.

https://github.com/McMartin/FRUT/blob/40e96f764e9139033cc1a3b3558ced4c81188a0d/Jucer2Reprojucer/main.cpp#L1436

McMartin:
Resolves #566.

@danra: could you please confirm that what I added to the README is correct?

### DIFF
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -1387,14 +1387,17 @@ int main(int argc, char* argv[])
       }
 
       const auto isAudioPlugin = projectType == "audioplug";
+      const auto pluginFormats = juce::StringArray::fromTokens(
+        jucerProject.getProperty("pluginFormats").toString(), ",", {});
       const auto hasJuceAudioProcessorsModule =
         jucerProject.getChildWithName("MODULES")
           .getChildWithProperty("id", "juce_audio_processors")
           .isValid();
 
       const auto hasVst2Interface = jucerVersionAsTuple > Version{4, 2, 3};
-      const auto isVstAudioPlugin =
-        isAudioPlugin && bool{jucerProject.getProperty("buildVST")};
+      const auto isVstAudioPlugin = isAudioPlugin
+                                    && (pluginFormats.contains("buildVST")
+                                        || bool{jucerProject.getProperty("buildVST")});
       const auto pluginHostVstOption = jucerProject.getChildWithName("JUCEOPTIONS")
                                          .getProperty("JUCE_PLUGINHOST_VST")
                                          .toString();
@@ -1415,8 +1418,9 @@ int main(int argc, char* argv[])
       }
 
       const auto supportsVst3 = exporterType == "XCODE_MAC" || isVSExporter;
-      const auto isVst3AudioPlugin =
-        isAudioPlugin && bool{jucerProject.getProperty("buildVST3")};
+      const auto isVst3AudioPlugin = isAudioPlugin
+                                     && (pluginFormats.contains("buildVST3")
+                                         || bool{jucerProject.getProperty("buildVST3")});
       const auto pluginHostVst3Option = jucerProject.getChildWithName("JUCEOPTIONS")
                                           .getProperty("JUCE_PLUGINHOST_VST3")
                                           .toString();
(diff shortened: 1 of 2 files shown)


Jucer2Reprojucer/main.cpp
```diff
@@ -1387,14 +1387,17 @@ int main(int argc, char* argv[])
...
         jucerProject.getChildWithName("MODULES")
           .getChildWithProperty("id", "juce_audio_processors")
           .isValid();
 
       const auto hasVst2Interface = jucerVersionAsTuple > Version{4, 2, 3};
-      const auto isVstAudioPlugin =
-        isAudioPlugin && bool{jucerProject.getProperty("buildVST")};
+      const auto isVstAudioPlugin = isAudioPlugin
+                                    && (bool{jucerProject.getProperty("buildVST")}
+                                        || pluginFormats.contains("buildVST"));
```
danra: I know you didn't ask for my code review but here it is anyway :)
It would be nice to DRY the strings here (and for the other formats)

README.rst
```diff
@@ -298,6 +298,9 @@ contributors:
       <a href="https://github.com/McMartin/FRUT/issues?q=is%3Aissue+author%3Adimitrisudell" title="Bug reports">🐛</a>
     </td>
     <td>
+      <a href="https://github.com/danra"><img src="https://github.com/danra.png" width="100"><br />Dan Raviv</a>
+      <br />
+      <a href="https://github.com/McMartin/FRUT/issues?q=is%3Aissue+author%3Adanra" title="Bug reports">🐛</a>
```
danra: 👍

Jucer2Reprojucer/main.cpp
danra: `hasProperty` can be used for more concise code instead of `getProperty`

danra: Probably makes more sense to check `pluginFormats` first to short-circuit most times, since that's where it would be for non-legacy jucer files

McMartin: > I know you didn't ask for my code review but here it is anyway :)

No problem at all. Reviews are always appreciated!


> It would be nice to DRY the strings here (and for the other formats)

Feel free to propose a concrete solution, since anything I could think of only adds more code 😅 


> `hasProperty` can be used for more concise code instead of `getProperty`

If the `"buildVST"` property is set to 0, `bool{getProperty("buildVST")}` will return `false`, while `hasProperty("buildVST")` will return `true`. So here, we can't use `hasProperty`.


> Probably makes more sense to check `pluginFormats` first to short-circuit most times, since that's where it would be for non-legacy jucer files

I totally agree! Addressed in 8460bc9c793a5a5ade961feec30b1103d346f344.

McMartin:
@danra it seems that GitHub doesn't let me "re-request a review" from you, but please have another look, thanks!

Jucer2Reprojucer/main.cpp
danra: > Feel free to propose a concrete solution, since anything I could think of only adds more code 😅
>
I suggest refactoring into a function which checks both, then it can just get the string once
 
> > `hasProperty` can be used for more concise code instead of `getProperty`
> 
> If the `"buildVST"` property is set to 0, `bool{getProperty("buildVST")}` will return `false`, while `hasProperty("buildVST")` will return `true`. So here, we can't use `hasProperty`.

Oops, right.

> > Probably makes more sense to check `pluginFormats` first to short-circuit most times, since that's where it would be for non-legacy jucer files
> 
> I totally agree! Addressed in [8460bc9](https://github.com/McMartin/FRUT/commit/8460bc9c793a5a5ade961feec30b1103d346f344).
👍

McMartin: > I suggest refactoring into a function which checks both, then it can just get the string once

That's actually what I had in mind, but I couldn't find a solution that was elegant enough. I'll leave the code as-is for now. I have plans to change how `Jucer2Reprojucer` works internally in the future anyway.

McMartin:
@danra thank you for the review! I'll merge this PR as soon as CI is green.

Issue #525, opened 2019-09-24
Allow finer granularity for precompiled lib paths in JUCE modules


TheSlowGrowth:
**Use case:**
A JUCE Module contains precompiled static libraries. It is included in a JUCE project that gets built on different IDE / compiler versions from the same source code.

**Problem:**
On Windows it is possible to have precompiled libraries for different versions of Visual Studio as well as Debug and Release versions in the JUCE module (simply because all these properties appear in the library search path inside the module - according to the [JUCE Module Format](https://github.com/WeAreROLI/JUCE/blob/master/modules/JUCE%20Module%20Format.txt#L97)). On all other platforms, there is basically just one path - making it impossible to work with multiple versions of XCode and forcing to link against release versions of the libraries even in debug builds. 

**Proposed solution:**
Maybe it would be possible to follow some logic like this (example for OSX):
````
libSearchPath = "${moduleDir}/libs/${XCodeVersion}/${arch}/${debug/release}/";
if not exists(libSearchPath)
   libSearchPath = "${moduleDir}/libs/${arch}";
````
This would stay compatible to the [JUCE Module Format](https://github.com/WeAreROLI/JUCE/blob/master/modules/JUCE%20Module%20Format.txt#L97) but also allow to work with multiple XCode versions as well as debug/release builds of the precompiled libraries.

**Random thoughts:**
I'm wondering why the [JUCE Module Format](https://github.com/WeAreROLI/JUCE/blob/master/modules/JUCE%20Module%20Format.txt#L97) even makes a difference between Windows and the other platforms. My best bet is that for Visual Studio, there are different exporters in Projucer, so the desired version is known right away. However, there is only one XCode exporter, making it much harder to determine which version of XCode will actually be used to open the generated project file. It probaby isn't possible to embed a "polymorphic" library search path in the XCode project file that leaves the actual path to be resolved by XCode when it actually opens the file, so maybe that's why we have only one single path for the precompiled libraries.
With CMake, this limitation is overcome because CMake knows what version of Xcode is installed.

McMartin:
Requires #517 to be resolved first.

PR #600, opened 2020-02-12
Add support for imported libraries.

sept-en:
Hi!
I tried to convert JUCE project which used a list of static libraries that were built separately by other scripts. The resulted CMake file was pretty okay, though I noticed that static libraries end up in EXTRA_LINKER_FLAGS resulting in incorrect CMake linking script (wrong linking order).

This PR adds support for such static libraries. CMake requires to mark these libraries as static imported libraries explicitly and then link them to the target.
I believe this PR solves #517, #525 (though it does not contain appropriate changes in `Jucer2Reprojucer`, which can be added once we finish with `Reprojucer.cmake` changes).

To use it, in exporter settings in FRUT **CMakeLists.txt** add another section with "IMPORTED_STATIC_LIBRARIES_PATHS" header that contains paths to appropriate static libraries:

```
jucer_export_target(
  "Linux Makefile"
  EXTRA_COMPILER_FLAGS
    "-I${CMAKE_CURRENT_SOURCE_DIR}/src"
    "-I${CMAKE_CURRENT_SOURCE_DIR}/src/local_third_party/boost"
  ...
  IMPORTED_STATIC_LIBRARIES_PATHS
    "${CMAKE_CURRENT_SOURCE_DIR}/src/third_party/boost/stage/lib/libboost_filesystem.a"
    "${CMAKE_CURRENT_SOURCE_DIR}/src/third_party/boost/stage/lib/libboost_iostreams.a"
    "${CMAKE_CURRENT_SOURCE_DIR}/src/third_party/boost/stage/lib/libboost_system.a"
    ...
)
```

I think this will also work for imported dynamic libs, then we can rename the setting to **"IMPORTED_LIBRARIES_PATHS"**, but I don't believe that this is the very popular case (and there are also some workaround ways for it).

As it is my very first time contributing to FRUT, your comments and suggestions are welcome, because something could be not FRUT-way.

Best regards,
Kirill.

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -745,6 +745,7 @@ function(jucer_export_target exporter)
     "EXTRA_COMPILER_FLAGS"
     "EXTRA_LINKER_FLAGS"
     "EXTERNAL_LIBRARIES_TO_LINK"
+    "IMPORTED_STATIC_LIBRARIES_PATHS"
   )
 
   if(exporter STREQUAL "Xcode (MacOSX)" OR exporter STREQUAL "Xcode (iOS)")
@@ -909,6 +910,10 @@ function(jucer_export_target exporter)
 
   if(DEFINED _EXTRA_LINKER_FLAGS)
     set(JUCER_EXTRA_LINKER_FLAGS "${_EXTRA_LINKER_FLAGS}" PARENT_SCOPE)
+  endif()
+
+  if(DEFINED _IMPORTED_STATIC_LIBRARIES_PATHS)
+      set(IMPORTED_STATIC_LIBRARIES_PATHS "${_IMPORTED_STATIC_LIBRARIES_PATHS}" PARENT_SCOPE)
   endif()
 
   if(DEFINED _EXTERNAL_LIBRARIES_TO_LINK)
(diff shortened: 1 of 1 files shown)


McMartin:
Hi Kirill,

Thanks a lot for using FRUT and for opening this pull request. I'm a bit confused by its contents though.

`Reprojucer.cmake` tries to reproduce (as its name suggests) what Projucer does. There is no "Imported Static Libraries Paths" setting in Projucer, so adding an `IMPORTED_STATIC_LIBRARIES_PATHS` keyword to `jucer_export_target` doesn't seem correct.

What exactly are you trying to achieve?
- If you have a `.jucer` project, which gets converted properly by `Jucer2Reprojucer`, but `Reprojucer.cmake` doesn't produce the same result as Projucer, then it's a bug in `Reprojucer.cmake` and I can work on fixing it once I know a bit more about your project.
- If you don't have a `.jucer` project, you can still use `Reprojucer.cmake`, but then you should add calls to the CMake commands that you want after the call to `jucer_project_end()` in your `CMakeLists.txt`.

sept-en:
Hi @McMartin , thank you for the review!

> Reprojucer.cmake tries to reproduce (as its name suggests) what Projucer does. There is no "Imported Static Libraries Paths" setting in Projucer, so adding an IMPORTED_STATIC_LIBRARIES_PATHS keyword to jucer_export_target doesn't seem correct.
I thought that the main goal of the project, according to README - _"FRUT makes it easy to build JUCE projects using CMake instead of Projucer."_ - isn't it?

For me, it was expectable that I can easily convert JUCE project to CMake via FRUT and slightly add any changes if needed keeping the generated FRUT structure (as it's similar to jucer project but in CMake format). It's really convenient and easy enough, thanks!
Nevertheless, when converting my jucer project it resulted in incorrect link script so I was unable to build an app.

In Projucer all extra libraries are specified in "Extra Linker Flags" setting. These are also are transferred via Reprojucer.cmake to CMake file, though that "just copy it from jucer part" is not enough for CMake - this way won't work for separately built static libraries (imported static libraries).

> If you have a .jucer project, which gets converted properly by Jucer2Reprojucer, but Reprojucer.cmake doesn't produce the same result as Projucer, then it's a bug in Reprojucer.cmake and I can work on fixing it once I know a bit more about your project.
Exactly. Described the wrong behavior above.

I also understand your intentions to keep the Reprojucer.cmake code clean and corresponding Projucer's project structure and I really appreciate that.

So another option I can suggest is to refactor the following way:
- remove "IMPORTED_STATIC_LIBRARIES_PATHS" keyword and keep using "EXTRA_LINKER_FLAGS" as it was before;
- make a check whether each linker flags is expected to be imported static library, if yes, then perform appropriate CMake operations which are needed and were added by this PR.

What do you think about that?

Best regards,
Kirill.

McMartin:
Hi Kirill,

> Nevertheless, convertation of my jucer project resulted in incorrect link script - so I was unable to build an app.

In that case, please create an issue (https://github.com/McMartin/FRUT/issues/new) describing in details what doesn't work. If you could attach your `.jucer` file, that would help me a lot to investigate the problem. 

Thanks!

McMartin:
Closing this PR, since there is no equivalent setting in Projucer.

PR #618, opened 2020-03-29
Use modern "--parallel" argument when building

McMartin:


McMartin:
Closing this PR because it doesn't work as expected...

PR #638, opened 2020-05-31
"Getting Started" should specify Juce modules root when directing user to run Reprojucer

importjingles:
"Getting Started" should specify Juce modules root when directing user to run Reprojucer

Otherwise an error is raised similar to:
error: The module juce_audio_basics requires a global path. You should pass the JUCE modules global path using --juce-modules.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,9 +154,11 @@ Then we convert ``MyGreatProject.jucer`` to a new ``CMakeLists.txt`` file: ::
 
   $ cd <root>/MyGreatProject/
 
-  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake
+  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake --juce-modules="../JUCE/modules"
 
   <root>/MyGreatProject/CMakeLists.txt has been successfully generated.
+
+*Note that the* ``--juce-modules`` *argument is only necessary if* ``MyGreatProject.jucer`` *uses modules from the global "JUCE Modules" path set in Projucer.*
 
 Now we can build ``MyGreatProject`` using CMake: ::
 


importjingles:
Added a more informative commit message and I'm updating this PR to be more informative

README.rst
```diff
@@ -154,7 +154,7 @@ Then we convert ``MyGreatProject.jucer`` to a new ``CMakeLists.txt`` file: ::
 
   $ cd <root>/MyGreatProject/
 
-  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake
+  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake --juce-modules="../JUCE/modules"
```
McMartin, line 157: Specifying `--juce-modules` is only necessary for `.jucer` projects that use JUCE modules from the Global Path set in Projucer. It will result in 
```
set(JUCE_MODULES_GLOBAL_PATH "${CMAKE_CURRENT_LIST_DIR}/../JUCE/modules")
```
being written in the `CMakeLists.txt` file, but `JUCE_MODULES_GLOBAL_PATH` won't be used anywhere if the `.jucer` project doesn't rely on the Global Path, so it might be quite confusing in that case.

I guess adding something like:
```diff
--- a/README.rst
+++ b/README.rst
@@ -158,6 +158,15 @@ Then we convert ``MyGreatProject.jucer`` to a new ``CMakeLists.txt`` file: ::

   <root>/MyGreatProject/CMakeLists.txt has been successfully generated.

+If ``MyGreatProject.jucer`` uses JUCE modules from the Global Path set in Projucer, we
+also need to pass ``--juce-modules="../JUCE/modules"`` to ``Jucer2Reprojucer`` ::
+
+  $ cd <root>/MyGreatProject/
+
+  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake --juce-modules="../JUCE/modules"
+
+  <root>/MyGreatProject/CMakeLists.txt has been successfully generated.
+
 Now we can build ``MyGreatProject`` using CMake: ::
```
would be an improvement, but it is quite verbose.

importjingles:
Ahh right, I understand. Yeah, I remember setting the global path in projucer. Sadly I'm not super familiar with juce and had other priorites, so I didn't feel like digging too hard to try and understand what was happening (sorry).

Yeah, I don't see a way around specifying the modules root in this case... especially if the jucer file was copied from machine to machine or you moved Juce for some reason.

README.rst
importjingles: Hmm, yeah, you could flip this around though so that it's a footnote rather than a "you must understand this first". Helps with the optic flow a bit and lowers the visual importance of the secondary information.

```suggestion
  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake --juce-modules="../JUCE/modules"

*Note that the ``--juce-modules`` parameter is only necessary if your jucer file uses modules from the Global Path set in Projucer.*
```

McMartin:
> I didn't feel like digging too hard to try and understand what was happening (sorry).

No problem at all! I'm super happy when people dig into their issues and give me answers/solutions, but I really don't mind if you don't.

It would actually be possible to read the XML file where Projucer stores the global paths. For instance, on Windows, that's `%APPDATA%\Projucer\Projucer.settings`. But you're totally right that we still need the `--juce-modules` argument, in case the `.jucer` file was moved to a machine where Projucer didn't run (or where the global path was not set).

README.rst
importjingles: I don't know how you made your nice inline diff...

importjingles:
> It would actually be possible to read the XML file where Projucer stores the global paths
I had thought of that too. But I didn't know where it stored it or how to articulate it (it's been a long day). Sounds like it could be worth doing that by default, but allow the user to override using --juce-modules.

I wonder then if this PR is really necessary.

McMartin:
> I wonder then if this PR is really necessary.

We could add something to the README now, then remove it once `Jucer2Reprojucer` can read the global paths from the Projucer settings. However, this is your PR, so it's your call.

README.rst
McMartin: To get syntax highlighting, you can specify a "language" right after the 3 back-ticks that start the code block. I used `diff`, and you used `suggestion`, which turns the code block into this "Suggested change" widget.

McMartin: I really like the "footnote" idea. I would suggest the following then:
```diff
@@ -154,10 +154,14 @@ Then we convert ``MyGreatProject.jucer`` to a new ``CMakeLists.txt`` file: ::

   $ cd <root>/MyGreatProject/

-  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake
+  $ ../FRUT/prefix/FRUT/bin/Jucer2Reprojucer MyGreatProject.jucer ../FRUT/prefix/FRUT/cmake/Reprojucer.cmake --juce-modules="../JUCE/modules"

   <root>/MyGreatProject/CMakeLists.txt has been successfully generated.

+*Note that the* ``--juce-modules`` *argument is only necessary if*
+``MyGreatProject.jucer`` *uses modules from the global "JUCE Modules" path set in
+Projucer.*
+
 Now we can build ``MyGreatProject`` using CMake: ::

   $ cd <root>/MyGreatProject/
```

(unfortunately, `*` and <tt>``</tt> cannot be nested in reST)

importjingles:
Please check my latest commit, which includes the argument description.

README.rst
```diff
@@ -154,10 +154,12 @@ Then we convert ``MyGreatProject.jucer`` to a new ``CMakeLists.txt`` file: ::
...
 
   <root>/MyGreatProject/CMakeLists.txt has been successfully generated.
 
+*Note that the* ``--juce-modules`` *argument is only necessary if* ``MyGreatProject.jucer`` *uses modules from the global "JUCE Modules" path set in Projucer.*
```
McMartin, line 161: Please break this line into several ones, so they don't go over 90 characters.
I know that many lines in this file don't respect that rule, but it's either to accommodate the reStructuredText syntax or to make inline HTML easier to read.

importjingles:
@McMartin I will get back to this Alain but I had some family business to attend to.

In the meantime, feel free to take ownership of this.

Take care bud.

McMartin:
Since https://github.com/McMartin/FRUT/pull/644, `Jucer2Reprojucer` can read JUCE and user modules global paths from Projucer's settings directly. I don't think this PR is necessary anymore, so I'm closing it.

PR #641, opened 2020-06-03
Improve error messages when global paths are required

McMartin:
Related to my comment there: https://github.com/McMartin/FRUT/pull/638#pullrequestreview-423895627

@importjingles, @MartyLake: please review, thanks!

### DIFF
--- a/Jucer2Reprojucer/main.cpp
+++ b/Jucer2Reprojucer/main.cpp
@@ -1322,12 +1322,14 @@ int main(int argc, char* argv[])
           printError("The module " + moduleName.toStdString()
                      + " requires a global path. You should pass the JUCE modules global "
                        "path using --juce-modules.");
+          return 1;
         }
         if (!isJuceModule && userModulesPath.isEmpty())
         {
           printError("The module " + moduleName.toStdString()
                      + " requires a global path. You should pass the user modules global "
                        "path using --user-modules.");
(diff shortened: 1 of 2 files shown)


Jucer2Reprojucer/main.cpp
```diff
@@ -1366,12 +1370,12 @@ int main(int argc, char* argv[])
       wLn("  PATH \"",
           useGlobalPath ? (isJuceModule ? "${JUCE_MODULES_GLOBAL_PATH}"
                                         : "${USER_MODULES_GLOBAL_PATH}")
-                        : relativeModulePath.replace("\\", "/"),
+                        : modulePath.replace("\\", "/"), //
```
importjingles, line 1373: Is the comment superfluous? Otherwise looks good :)

McMartin: Very good catch! I added that comment to track down the origin of the `relativeModulePath` variable using `gitk`'s "Origin of this line" feature. I simply forgot to remove it before making the commit :sweat_smile:

McMartin: Addressed in 0547d00b8762d96ec83194ccc5af1e3305279994.

McMartin:
@importjingles Please have a look at the last commit and let me know whether I should change anything. Thanks!

PR #645, opened 2020-06-21
Introduce Jucer2CMake

McMartin:
It will first replace Jucer2Reprojucer, then in the near future, it will
support converting .jucer project files into CMakeLists.txt files that
use the CMake API of JUCE 6.


MartyLake:
Lgtm!

PR #655, opened 2020-10-26
Write configuration names that are valid for CMake

### LINKED ISSUE #520: Certain configuration names cause CMake errors

McMartin:
Closes #520 and #521.


McMartin:

McMartin:
@MartyLake gentle ping

Jucer2CMake/src/main.cpp
```diff
@@ -181,6 +182,19 @@ juce::String makeValidIdentifier(juce::String s)
 }
 
 
+// Matches _FRUT_make_valid_configuration_name in Reprojucer.cmake.
+juce::String makeValidConfigurationName(const juce::String& configurationName)
+{
+  auto validName = configurationName.toStdString();
+
+  validName = std::regex_replace(validName, std::regex{"[^A-Za-z0-9_]+"}, " ");
```
MartyLake, line 190: maybe extract regex already duplicated once ?

MartyLake (approved):
LGTM, with minor suggestion you can ignore :)

Jucer2CMake/src/main.cpp
McMartin: `^[A-Za-z0-9_]+$` and `[^A-Za-z0-9_]+` are not the same.

PR #676, opened 2021-03-06
Handle special module config flags

McMartin:


Jucer2CMake/src/juce6.hpp
```diff
@@ -109,6 +109,15 @@ inline void writeJuce6CMakeLists(const Arguments&, const juce::XmlElement& jucer
 
     wLn(juceAddFunction, "(", targetName);
 
+    bool needEmptyLine = false;
```
MartyLake, line 123: What’s the advantage of having this variable with a side-effect compared to emitting "wLn();" at each "needEmptyLine = true;" ?

MartyLake (approved):
LGTM !

Jucer2CMake/src/juce6.hpp
McMartin: If we had `wLn();` instead of `needEmptyLine = true;`, then we would get
```cmake
juce_add_console_app(consoleapp6
  VERSION "0.1"
  COMPANY_NAME "The Name of the Company"
  COMPANY_COPYRIGHT "The Copyright of the Company"
  COMPANY_WEBSITE "The Website of the Company"
  COMPANY_EMAIL "The E-mail of the Company"




  NEEDS_CURL TRUE
)
```

But since it was not clear, I guess I need to rewrite some commits.

McMartin: I made the intent clearer by moving that commit at the end of the branch. Please have another look, thanks!

MartyLake: Alright ! it’s easier to see the sideeffect with this commit ordering !

PR #681, opened 2021-04-13
Write "companyName" field in JucerHeader.h like Projucer does

### LINKED ISSUE #680: Company name missing from JuceHeader
Currently the ProjectInfo::companyName field doesn't get added from the Jucer file, when the JuceHeader is generated through FRUT.

McMartin:
Resolves #680.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -342,6 +342,9 @@ contributors:
       <a href="https://github.com/McMartin/FRUT/issues?q=is%3Aissue+author%3AE-Fir" title="Bug reports">🐛</a>
     </td>
     <td>
+      <a href="https://github.com/benediktsailer"><img src="https://github.com/benediktsailer.png" width="100"><br />Benedikt Sailer</a>
+      <br />
+      <a href="https://github.com/McMartin/FRUT/issues?q=is%3Aissue+author%3Abenediktsailer" title="Bug reports">🐛</a>
     </td>
   </tr>
   </tbody>
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -4157,6 +4157,14 @@ function(_FRUT_generate_JuceHeader_header)
     string(APPEND modules_includes "\n")
   endif()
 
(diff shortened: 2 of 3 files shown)


benediktsailer:
I think there's a small type-o in `cmake/data/JuceHeader.h.in`

It says
`const char* const projectName = "@JUCER_PROJECT_NAME@";@company_name_field@`

on line 26 where it should say 
`
const char* const projectName = "@JUCER_PROJECT_NAME@"; 
const char* const companyName = "@company_name_field@";
`
 instead

McMartin:
@benediktsailer see the changes in `cmake/Reprojucer.cmake` for what `@company_name_field@` will be replaced by. It's more than just the company name. Maybe I should rename `company_name_field` to `company_name_line`.

benediktsailer:
I see! My bad! :)

McMartin:
@benediktsailer No problem. Please do test and let me know whether it works as expected. Thanks!

benediktsailer:
Working like a charm! Thank you!

MartyLake:
If you want to strictly follow the change of behavior, then this PR LGTM as-is.

However, versioning the API change like that is hard to follow if one is not in this exact commit frame but only reviewing one file at a time (in it's text editor for example). Since companyName is not a breaking API change, maybe only adding is less complex and works equally well ? (untested)

```diff
#if ! JUCE_DONT_DECLARE_PROJECTINFO
namespace ProjectInfo
{
    const char* const  projectName    = "@JUCER_PROJECT_NAME@";
+   const char* const  companyName    = "@JUCER_COMPANY_NAME@";
    const char* const  versionString  = "@JUCER_PROJECT_VERSION@";
    const int          versionNumber  = @JUCER_PROJECT_VERSION_AS_HEX@;
}
```

McMartin:
Always having `ProjectInfo::companyName` should indeed work, but `Reprojucer.cmake` tries to reproduce the Projucer behavior as closely as possible. Making the code more complex to achieve that is fine.

PR #687, opened 2021-06-06
Fix handling settings that contain paths relative to the build folder

### LINKED ISSUE #686: EXTRA_CUSTOM_FRAMEWORKS - different path needed for CMake, but wrong in Xcode
Hey FRUT devs,

thanks for this project, we really need the ability to build our desktop app via command line / CI.

We have a problem with the EXTRA_CUSTOM_FRAMEWORKS paths. We have custom c++ libs which we are using as frameworks in Xcode and Xcode expects to have a path which are relativ to the xcodeproj file. But when building with cmake, the paths have to be different, because it needs to be relativ to the CMakeLists.txt. This is only because the cmake build checks if the paths existing. Projucer is not doing that, the content of "Extra Custom Frameworks" is only copied to the xcode project file.

Is it possible to switch off the check of whether the path exists for the cmake build? 
or how can I configure it to able to build with cmake and have the path correct for xcode?

Thanks
David

McMartin:
Resolves #686 


MartyLake (approved):
LGTM! :)

PR #701, opened 2021-10-29
Remove Jucer2Reprojucer

McMartin:
It was deprecated in 1bce3e5f8ddfaf91f695220694c123f6195c7ef7 (June
2020), whose commit message reads

    Mark Jucer2Reprojucer as deprecated

    We will keep it in the repository for about 6 months and remove it at
    the beginning of 2021.

so it is really time to remove it.


McMartin:
@MartyLake: FYI. No need to review, but you might appreciate the diff stats 😄

PR #665, opened 2021-01-04
Support macOS 11.0 SDK

scotchi:
Pretty trivial request to make things work with the new macOS 11 SDKs.  I'm not sure if it'd make sense to support as of yet unreleased 11.x versions in the same go...

PR #725, opened 2022-07-02
Rename settings from "OSX" to "macOS" and support recent macOS versions

McMartin:
Supersedes #665


cmake/Reprojucer.cmake
```diff
@@ -1804,7 +1804,10 @@ function(jucer_export_target_configuration
     if(version MATCHES "^10\\.([5-6]) SDK$"
         AND DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5.3.2)
       set(JUCER_OSX_BASE_SDK_VERSION_${config} "10.${CMAKE_MATCH_1}" PARENT_SCOPE)
-    elseif(version MATCHES "^10\\.([7-9]|1[0-5]) SDK$")
+    elseif(version MATCHES "^10\\.([7-9]|10) SDK$"
+        AND DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 5.4.4)
+      set(JUCER_OSX_BASE_SDK_VERSION_${config} "10.${CMAKE_MATCH_1}" PARENT_SCOPE)
+    elseif(version MATCHES "^10\\.(1[1-5]) SDK$")
```
MartyLake: For regex-noob like me, those regexes are hard to decipher. I guess it means

* from 10.7 to 10.10

and

* from 10.11 to 10.15

but I couldn't say if there is a bug here or not.

MartyLake: maybe VERSION_LESS would be more readable, but unfortunately too verbose I guess.

MartyLake: oh wait, but then you cannot do "CMAKE_MATCH_1" anymore.

MartyLake (approved):
LGTM

cmake/Reprojucer.cmake
McMartin: > For regex-noob like me, those regexes are hard to decipher. I guess it means
> 
>     * from 10.7 to 10.10
> 
> 
> and
> 
>     * from 10.11 to 10.15
> 
> 
> but I couldn't say if there is a bug here or not.

You got the regex right.
Older SDKs (10.7 to 10.10) are only supported by older Projucer (before 5.4.4)

> maybe VERSION_LESS would be more readable, but unfortunately too verbose I guess.

Using `VERSION_LESS` would require stripping ` SDK` first, which would make the code more complex IMHO.

Issue #726, opened 2022-07-15
Issue with "Header Search Paths" and "Extra Library Search Paths" in exporter


Riuzakiii:
Hi,

First of all, thanks for maintaining this project ! Here is my issue :

The project I'm working on uses relative paths for the "Header Search Paths" and "Extra Library Search Paths" in the exporter settings of Projucer. Switching to FRUT, the headers added using the "Header Search Paths" were unfound. 

The reason seems to be the following
1.  _Jucer2CMake_ converts the relative paths of "Header Search Paths" and "Extra Library Search Paths" from relative to the target project directory, to relative to the .jucer directory.
With this folder structure

root/
├─ MyProject/
│  ├─ Builds/
│  │  ├─ TargetFolder/
│  ├─ MyProject.jucer

Accessing a file in root would look like "../../../myfile.h", but _Jucer2CMake_ transforms it to "../myfile.h" (relative to the .jucer)

2. _Reprojucer.cmake_ uses the `_FRUT_abs_path_based_on_jucer_target_project_folder` function, which fails to give the correct path because it is now relative to the .jucer directory
This does not happen with "Extra Library Search Paths" because _Reprojucer.cmake_ uses the `_FRUT_abs_path_based_on_jucer_project_dir` function.

To fix this, I switched both "Header Search Paths" and "Extra Library Search Paths" to fully use paths relative to the target project directory, which will make the _.jucer_ and _CMakeLists.txt_ match to avoid confusion. You can find my changes here https://github.com/Riuzakiii/FRUT/commit/ed2db9052778a616f598f09246b5d147e10de625. 

Do you think this would be a good fix ?

McMartin:
Hi @Riuzakiii,

Thanks a lot for reporting this bug! 

In https://github.com/McMartin/FRUT/pull/687, I tried to fix the issues related to settings that contain paths relative to the target project directory, but it seems that I missed these...

I left a few comments on your changes: they go in the right direction, but would need a bit of polish to be merged. Feel free to open a pull request though! (I would recommend using another branch than `main` in your repository)

Riuzakiii:
Thanks, I made changes based on your comments and switched to using a `develop` branch instead of `main`. Here's the pull request : https://github.com/McMartin/FRUT/pull/727

PR #727, opened 2022-07-16
Fix HEADER_SEARCH_PATHS and EXTRA_LIBRARY_SEARCH_PATHS

Riuzakiii:
They were relative to the jucer folder instead of project folder

Fixes https://github.com/McMartin/FRUT/issues/726

### DIFF
--- a/Jucer2CMake/src/reprojucer.hpp
+++ b/Jucer2CMake/src/reprojucer.hpp
@@ -2,6 +2,7 @@
 // Copyright (C) 2017-2022  Alain Martin
 // Copyright (C) 2017  Florian Goltz
 // Copyright (C) 2019  Johannes Elliesen
+// Copyright (C) 2022  Thiébaud Fuchs
 //
 // This file is part of FRUT.
 //
@@ -157,6 +158,22 @@ inline juce::StringArray parsePreprocessorDefinitions(const juce::String& input)
   }
 
   return definitions;
+}
+
+
+inline juce::StringArray parseSearchPaths(const juce::String& input)
+{
+  juce::StringArray unquotedPaths;
+
+  for (const auto& path : juce::StringArray::fromTokens(input, ";\r\n", {}))
+  {
+    if (path.isNotEmpty())
+    {
+      unquotedPaths.add(path.unquoted());
+    }
+  }
+
+  return unquotedPaths;
 }
 
 
@@ -758,10 +775,8 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
 
     convertSettingAsListIfDefined(jucerProject, "defines", "PREPROCESSOR_DEFINITIONS",
                                   parsePreprocessorDefinitions);
-    convertSettingAsListIfDefined(
-      jucerProject, "headerPath", "HEADER_SEARCH_PATHS", [](const juce::String& value) {
-        return juce::StringArray::fromTokens(value, ";\r\n", {});
-      });
+    convertSettingAsListIfDefined(jucerProject, "headerPath", "HEADER_SEARCH_PATHS",
+                                  parseSearchPaths);
 
     convertSettingIfDefined(jucerProject, "postExportShellCommandPosix",
                             "POST_EXPORT_SHELL_COMMAND_MACOS_LINUX", {});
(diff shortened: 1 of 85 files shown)


Jucer2CMake/src/reprojucer.hpp
```diff
@@ -507,6 +507,9 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
       }
     };
 
+  const auto convertSearchPaths = [](const juce::String& searchPaths) -> juce::StringArray
```
McMartin, line 510: Since there is no capture, this can be a free function, like `parsePreprocessorDefinitions`. I would call it `parseSearchPaths` since they have a similar role.

Jucer2CMake/src/reprojucer.hpp
```diff
@@ -2068,39 +2069,6 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
...
-          juce::StringArray absOrRelToJucerFileDirPaths;
-
-          for (const auto& path : juce::StringArray::fromTokens(searchPaths, ";\r\n", {}))
-          {
-            if (path.isEmpty())
-            {
-              continue;
-            }
-
-            const auto unquotedPath = path.unquoted();
```
McMartin, line 2088: This line is important and should not be lost. Please preserve it in the new function.

Jucer2CMake/src/reprojucer.hpp
```diff
@@ -2058,49 +2059,10 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
...
         convertSettingIfDefined(configuration, "binaryPath", "BINARY_LOCATION", {});
 
-        const auto isAbsolutePath = [](const juce::String& path) {
-          return path.startsWithChar('/') || path.startsWithChar('~')
-                 || path.startsWithChar('$')
-                 || (juce::CharacterFunctions::isLetter(path[0]) && path[1] == ':');
-        };
-
         const auto jucerFileDir = jucerFile.getParentDirectory();
         const auto targetProjectDir =
```
McMartin, line 2063: `jucerFileDir` and `targetProjectDir` also become unused and can be removed.

cmake/Reprojucer.cmake
```diff
@@ -1688,13 +1688,7 @@ function(jucer_export_target_configuration
...
 
   if(DEFINED _EXTRA_LIBRARY_SEARCH_PATHS)
-    set(library_search_paths "")
-    foreach(path IN LISTS _EXTRA_LIBRARY_SEARCH_PATHS)
-      file(TO_CMAKE_PATH "${path}" path)
-      _FRUT_abs_path_based_on_jucer_project_dir(path "${path}")
-      list(APPEND library_search_paths "${path}")
-    endforeach()
-    set(JUCER_EXTRA_LIBRARY_SEARCH_PATHS_${config} "${library_search_paths}" PARENT_SCOPE)
+    set(JUCER_EXTRA_LIBRARY_SEARCH_PATHS_${config} "${_EXTRA_LIBRARY_SEARCH_PATHS}" PARENT_SCOPE)
```
McMartin, line 1691: Now that `EXTRA_LIBRARY_SEARCH_PATHS` works like `HEADER_SEARCH_PATHS`, there should be a bit of code related to it below
https://github.com/McMartin/FRUT/blob/b874ee48365bfdfaaba029a1a452632e66434071/Jucer2CMake/src/reprojucer.hpp#L1410-L1413

McMartin (commented):
Thanks a lot for opening this pull request!

I found a few things that I either missed or just didn't think about earlier. Let me know if something is not clear. I'm also happy to push fixup commits if you want.

Riuzakiii:
I updated the PR based on your comments, hope I did not miss too much this time ! the pipeline is failing because of the changes in the generated CMakeLists.txt, but I don't think I can get the artefacts from the pipeline, you probably have to do that

Riuzakiii:
Btw I did not format the code because it was also changing other parts of the code, the CI does not seem to check this file either

McMartin:
@Riuzakiii Thanks a lot for addressing my comments quickly! 

> the pipeline is failing because of the changes in the generated CMakeLists.txt,

Indeed, I pushed 74905a8b2f8ddd3f05ffb0e254b6b443af104ced to address that (please keep it as a separate commit).

> the CI does not seem to check this file either

Very good catch ! 👏🏻 
I opened https://github.com/McMartin/FRUT/pull/728 to fix it. I'll merge it first, so you'll have to rebase on top of the latest `main`.

Riuzakiii:
No problem, I don't mind at all ! I will rebase the branch after you merge https://github.com/McMartin/FRUT/pull/728 and keep the changes of the generated file as a separate commit.

Riuzakiii:
Looks good to me, thanks for adding me to the table of contributors

PR #731, opened 2022-07-21
Set Xcode "ARCH" settings like Projucer does

### LINKED ISSUE #700: ARM support
Currently building a JUCE project through FRUT ignores the ARM build target. 

On Xcode I can build my Projucer build configuration for "All Apple Devices", so it compiles a universal binary for x86_64 and arm64. 

I would expect FRUT to do the same if not otherwise specified with an `ARCH=x86_64` flag in the Custom XCode Flags field on Projucer, which is how you make XCode ignore the arm build target.

McMartin:
Resolves #700.

### DIFF
--- a/Jucer2CMake/src/reprojucer.hpp
+++ b/Jucer2CMake/src/reprojucer.hpp
@@ -1677,6 +1677,11 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
             });
         }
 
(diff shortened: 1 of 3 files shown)


McMartin:
The commits of this PR were mistakenly part of #732, which was merged to `main`. Since that was not the desired outcome, I've hard-reset `main` to 153d6bc00977fb9b6e9d8a6c0b1c507b37690aa9 and I've opened #734 to merge these changes in due time.

PR #734, opened 2022-07-23
Set Xcode `*ARCH*` settings like Projucer does

### LINKED ISSUE #700: ARM support

McMartin:
Resolves #700.

### DIFF
--- a/cmake/Reprojucer.cmake
+++ b/cmake/Reprojucer.cmake
@@ -5777,8 +5777,15 @@ function(_FRUT_set_compiler_and_linker_settings_APPLE target)
 
   if(NOT IOS AND NOT (DEFINED JUCER_VERSION AND JUCER_VERSION VERSION_LESS 6.0.2))
     unset(valid_archs)
+    set(excluded_archs "")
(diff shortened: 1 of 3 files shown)


McMartin:

cmake/Reprojucer.cmake
```diff
@@ -5739,6 +5739,12 @@ function(_FRUT_set_compiler_and_linker_settings_APPLE target)
           set_target_properties(${target} PROPERTIES
             XCODE_ATTRIBUTE_ARCHS[variant=${config}] "${JUCER_XCODE_ARCHS_${config}}"
           )
+        else()
+          if(NOT CMAKE_VERSION VERSION_LESS 3.18.1 AND NOT XCODE_VERSION VERSION_LESS 12)
+            set_target_properties(${target} PROPERTIES
```
MartyLake, line 5754: good catch on that one

MartyLake (approved):
LGTM

Issue #749, opened 2022-12-08
Error during cmake --build . --target install ("JUCE requires c++17 or later")


lucreciaef:
Hi! Long story short, I have an assignment in which we were asked to export a project from Projuicer to Visual Studio. I would like to use Clion instead, but that requires that I convert the .jucer file to .cmake

I'm using Windows10, so I installed WSL Ubuntu, got the whole thing running (after 3 hours of trial and error)
I installed all the dependencies I could, but when running
`$ cmake --build . --target install` 
the console says "Building CXX object Jucer2Make......" and then after 5 lines it breaks saying 
"91 | #error "JUCE requires C++17 or later"

I have the latest version of g++ and clang installed on the ubuntu vm. I installed g++-multilib too.

Do I need to add something additional to the build/target command ? I am learning C++ and I'm quite new to linux, so I apologize if this is a complete noob question. I've exhausted my options.

In the meantime I can continue working on Visual Studio but it's inconvenient once I'm already used to my Clion IDE :(

Thanks in advance to anyone who could give me a hint

McMartin:
Hi @lucreciaef,

On Windows, you can use CLion with the MSVC toolchain that comes with Visual Studio. See https://www.jetbrains.com/help/clion/quick-tutorial-on-configuring-clion-on-windows.html#MSVC. You don't have to use WSL.

However, if you still want to use WSL, FRUT should work just fine. Could you please share the `CMakeCache.txt` file in a new comment? I'd like to see how FRUT was configured.

cuanyu:
I am having the same problem on my Ubuntu 20.04 machine too when I try build and install FRUT. 

> /home/tom/JUCE/modules/juce_core/system/juce_CompilerSupport.h:91:3: error: "JUCE requires C++17 or later"
 #error "JUCE requires C++17 or later"

The JUCE itself could be correctly built stand alone. I am using gcc 9.4.0 and clang 10.0.0-4ubuntu1.

lucreciaef:
Hi @McMartin 
Thank you very much for the quick reply.

I was fortunately able to make this work today! So I'll share what happened here for the record:
I initially tried with Ubuntu so that I could follow the exact installation instructions in the FRUT readme, which are linux commands. It doesn't work otherwise (or I'm not aware of how to do it). So, I wasn't able to create the `CMakeLists.txt` file myself - a colleague with a Linux machine generated it for me, and I pasted it into my project file and changed the folder paths.

[...]

PR #750, opened 2022-12-09
Support building FRUT with JUCE 7.0.3

McMartin:
Resolves https://github.com/McMartin/FRUT/issues/749


McMartin:

Jucer2CMake/CMakeLists.txt
```diff
@@ -71,7 +66,10 @@ target_compile_definitions(Jucer2CMake_JUCE PUBLIC
 )
 
 if(APPLE)
-  target_compile_options(Jucer2CMake_JUCE PRIVATE -Wno-deprecated-declarations)
+  target_compile_options(Jucer2CMake_JUCE PRIVATE
+    -Wno-deprecated-declarations
+    -Wno-register
```
MartyLake, line 71: What is -Wno-register with C++17 ? a new unnecessary warning generated by setting cxx17 ?

MartyLake (approved):
apart from minor question, LGTM

Jucer2CMake/CMakeLists.txt
McMartin: When building older versions of JUCE with C++17 we get:
```
error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
```
from third-party libraries included in JUCE (jpglib and zlib). There's nothing much we can do but disable that warning.

PR #759, opened 2023-05-28
Fix generating shell scripts with variables

aostrowski:
JUCE projects allow using shell script, for instance when defining post
build actions. When the shell script referred to variables using braced
notation, e. g. ${PULP}, FRUT made CMake resolve such a variable at
build system generation time instead of at build time.

For instance, given the following Post-Build Shell Script:

```
VAR="It works!"
echo "${VAR}"
```

we will end up with a postbuild.sh file with the following content:

```
VAR="It works!"
echo ""
```

As CMake will recognize `${VAR}` as a variable reference and will
immediately resolve it. The following script would works fine:

```
VAR="It works!"
echo "$VAR"
```

but the former is still a properly written shell script.

This commit treats the shell scripts as bracket arguments instead of
quoted arguments to fix this issue.

### DIFF
--- a/Jucer2CMake/src/reprojucer.hpp
+++ b/Jucer2CMake/src/reprojucer.hpp
@@ -3,6 +3,7 @@
 // Copyright (C) 2017  Florian Goltz
 // Copyright (C) 2019  Johannes Elliesen
 // Copyright (C) 2022  Thiébaud Fuchs
+// Copyright (C) 2023  Adrian Ostrowski
 //
 // This file is part of FRUT.
 //
@@ -387,6 +388,18 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
     }
   };
 
+  const auto writeBracket = [&wLn](const juce::String& cmakeKeyword,
+                                   const juce::String& value) {
+    if (value.isEmpty())
+    {
+      wLn("  # ", cmakeKeyword);
+    }
+    else
+    {
+      wLn("  ", cmakeKeyword, " [=[", value, "]=]");
+    }
+  };
+
   const auto writeQuoted = [&wLn](const juce::String& cmakeKeyword,
                                   const juce::String& value) {
     if (value.isEmpty())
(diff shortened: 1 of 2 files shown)


aostrowski:
Looks like the checks failed due to infra issues like: 

```
##[warning]An image label with the label macOS-10.15 does not exist.
##[error]The remote provider was unable to process the request.
```

@McMartin please advise.

Jucer2CMake/src/reprojucer.hpp
```diff
@@ -375,6 +375,18 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
 
   LineWriter wLn{outputStream};
 
+  const auto writeBracket = [&wLn](const juce::String& cmakeKeyword,
```
McMartin, line 378: I'd like the `write*` lambdas to be in the order of "least-complex" to "most-complex". With the addition of `writeBracket`, it would be:
- `writeUnquoted`
- `writeBracket`
- `writeQuoted`
- `writeList`

So please move `writeBracket` right after `writeUnquoted`.

Jucer2CMake/src/reprojucer.hpp
```diff
@@ -431,6 +443,17 @@ inline void writeReprojucerCMakeLists(const Arguments& args,
       writeQuoted(cmakeKeyword, value);
     };
 
+  const auto convertRawStringSetting =
```
McMartin, line 446: The `convert*Setting` lambdas also have a specific order. Please put `convertRawStringSetting` and `convertRawStringSettingIfDefined` between `convertOnOffSettingWithDefault` and `convertSettingAsList`.

McMartin (commented):
@aostrowski Thanks a lot for opening this pull request! I took care of fixing the CI builds in #760. Please apply my suggestions and rebase this PR on top of the latest `main`. Then I'll be more than happy to merge it!

Jucer2CMake/src/reprojucer.hpp
McMartin: Addressed in 22df5395c361c114111f985a2172a8ace1563de4.

Jucer2CMake/src/reprojucer.hpp
McMartin: Addressed in 22df5395c361c114111f985a2172a8ace1563de4.

Issue #780, opened 2025-03-12
_FRUT_version_to_dec miscalculating for AU builds with larger revision numbers.


TheAudioGoblin:
## Description:

When generating AU .component packages I encountered an issue where the version in info.plist -> AudioComponents was too high of a number, resulting in the .component becoming undetectable in auval. This issue only occurred for me with higher revision numbers i.e. >100.

## Example:

A plugin of version 1.1.1.108 had a version output in info.plist -> AudioComponents of 1157425104234217472 - note the correct calculation should be 16843116.

## Root Cause:

in Reproducer.cmake, _FRUT_version_to_dec():

``` 
math(EXPR dec_value "(${major} << 16) + (${minor} << 8) + ${patch}")
if(segments_size GREATER 3)
  list(GET segments 3 revision)
  math(EXPR dec_value "${dec_value} << 8 + ${revision}")
endif()
 ```

Due to the order of operations here, for a higher revision number i.e. 108, the dec_value becomes shifted by (8+108), resulting too large of an output.

## Proposed Solution:

Use bitwise OR operation instead of addition:
```
math(EXPR dec_value "(${major} << 16) | (${minor} << 8) | ${patch}")
if(segments_size GREATER 3)
  list(GET segments 3 revision)
  math(EXPR dec_value "${dec_value} << 8 | ${revision}")
endif()
```

This produces a more reliable calculation of the decimal value of a given version number.

PR #781, opened 2025-03-12
Fixes Issue #780: Replaced version component addition with bitwise OR

TheAudioGoblin:
- Replaced addition (`+`) with bitwise OR (`|`) in `_FRUT_version_to_dec` to correctly pack version components.
- Fixes incorrect results caused by addition-related overflow or bit corruption.
- Ensures reliable encoding of version numbers for all valid inputs.


MartyLake:
Hey @McMartin , I guess that the CI errors are not related to TheAudioGoblin changes. I tried to take a look , maybe the macOS vm there were updated (or not because `macOS-13` is still listed in https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml ).

Do you think you could take a look at some point, or maybe we could merge this one as-is since it is a simple change ?

